PR #17, opened 2020-08-16
Update istio images for istio 1.6.8

brendarearden:


images-list
```diff
@@ -1,14 +1,10 @@
-istio/sidecar_injector rancher/istio-sidecar_injector 1.5.9
```
cbron, line 1: You can leave all the old ones because we will need them for more 1.5.x releases

images-list
```diff
@@ -1,14 +1,10 @@
...
-istio/galley rancher/istio-galley 1.5.9
-istio/mixer rancher/istio-mixer 1.5.9
-istio/pilot rancher/istio-pilot 1.5.9
-istio/citadel rancher/istio-citadel 1.5.9
-istio/kubectl rancher/istio-kubectl 1.5.9
-istio/proxyv2 rancher/istio-proxyv2 1.5.9
-istio/node-agent-k8s rancher/istio-node-agent-k8s 1.5.9
+istio/mixer rancher/istio-mixer 1.6.8
+istio/pilot rancher/istio-pilot 1.6.8
+istio/coredns-plugin rancher/istio-coredns-plugin 0.2-istio-1.1
```
cbron, line 3: This one exists below

images-list
```diff
@@ -1,14 +1,10 @@
-istio/sidecar_injector rancher/istio-sidecar_injector 1.5.9
-istio/galley rancher/istio-galley 1.5.9
-istio/mixer rancher/istio-mixer 1.5.9
-istio/pilot rancher/istio-pilot 1.5.9
-istio/citadel rancher/istio-citadel 1.5.9
-istio/kubectl rancher/istio-kubectl 1.5.9
```
cbron, line 6: Is there no kubectl in the manifests ?

brendarearden: From what I saw in the manifests, kubectl was only used in the addons, specifically for grafana dashboard.

PR #19, opened 2020-08-17
add logging images

paynejacob:


images-list
```diff
@@ -34,3 +34,8 @@ us.gcr.io/k8s-artifacts-prod/external-dns/external-dns rancher/kubernetes-extern
 openpolicyagent/gatekeeper rancher/opa-gatekeeper v3.1.0-beta.11
 k8s.gcr.io/k8s-dns-node-cache rancher/k8s-dns-node-cache 1.15.7
 sonobuoy/sonobuoy rancher/sonobuoy-sonobuoy v0.16.3
+banzaicloud/logging-operator rancher/logging-operator 3.4.0
```
cbron, line 37: We like to prepend the original org to the image name -> `rancher/banzaicloud-logging-operator`
Also, any image that doesn't currently exist in our dockerhub will need to get a repo set up for it, I can give you details. Ideally that would get checked on PR CI checks, made issue for that.

images-list
```diff
@@ -34,3 +34,8 @@ us.gcr.io/k8s-artifacts-prod/external-dns/external-dns rancher/kubernetes-extern
...
+jimmidyson/configmap-reload rancher/configmap-reload v0.2.2
+paynejacob/fluent-bit-out-syslog rancher/fluent-bit-out-syslog latest
+banzaicloud/fluentd rancher/fluentd v1.10.4-alpine-2
```
cbron, line 40: We already have a fluentD repo but its custom, so this one should also be org-namespaced. It looks like banzai's is different image entirely than the official fluentD ? 
* https://hub.docker.com/_/fluentd?tab=tags
* https://hub.docker.com/r/banzaicloud/fluentd/tags
* https://hub.docker.com/r/rancher/fluentd/tags

images-list
```diff
@@ -34,3 +34,8 @@ us.gcr.io/k8s-artifacts-prod/external-dns/external-dns rancher/kubernetes-extern
 openpolicyagent/gatekeeper rancher/opa-gatekeeper v3.1.0-beta.11
 k8s.gcr.io/k8s-dns-node-cache rancher/k8s-dns-node-cache 1.15.7
 sonobuoy/sonobuoy rancher/sonobuoy-sonobuoy v0.16.3
+banzaicloud/logging-operator rancher/logging-operator 3.4.0
+jimmidyson/configmap-reload rancher/configmap-reload v0.2.2
+paynejacob/fluent-bit-out-syslog rancher/fluent-bit-out-syslog latest
```
cbron, line 39: If this image doesn't actually exist anywhere, then lets put version at 0.1.0

images-list
paynejacob: They may be compatible but this image has all the plugins installed that the operator supports so I think it is worth having 2.

images-list
paynejacob: Changed to 0.1.0, this image can be dropped once we have our own ci but until that is ready we can mirror my build.

paynejacob:
https://github.com/rancherlabs/eio/issues/200

PR #23, opened 2020-08-25
Add images for istio 1.7.0

brendarearden:
Add images for istio 1.7.0


cbron:
@brendarearden can you double check the git diff between last 1.6.x and 1.7.0 and make sure none of the other images changed

brendarearden:
https://github.com/istio/istio/compare/1.6.8...1.7.0 (having an issue loading because there were so many changes)
I went ahead and downloaded the zips and compared the manifets and also installed istioctl 1.7.0 and did a demo profile dump and based on that it looks like the grafana had a version changed and then the rest that use the global tag updated to 1.7.0

PR #26, opened 2020-08-28
Add k3s images

brandond:
I believe these images were previously pushed manually by @cjellick

### DIFF
--- a/images-list
+++ b/images-list
@@ -46,3 +46,6 @@ directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-ad
 bats/bats rancher/bats-bats v1.1.0
 curlimages/curl rancher/curlimages-curl 7.70.0
 busybox rancher/busybox 1.31.1
+coredns/coredns rancher/coredns-coredns 1.6.9
+k8s.gcr.io/metrics-server/metrics-server rancher/metrics-server v0.3.7
+k8s.gcr.io/pause rancher/pause 3.1


brandond:
OK, merging for now - multiarch for these is needed but can wait on the other PR.

images-list
```diff
@@ -46,3 +46,6 @@ directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-ad
 bats/bats rancher/bats-bats v1.1.0
 curlimages/curl rancher/curlimages-curl 7.70.0
 busybox rancher/busybox 1.31.1
+coredns/coredns rancher/coredns-coredns 1.6.9
+k8s.gcr.io/metrics-server/metrics-server rancher/metrics-server v0.3.7
```
deniseschannon, line 50: why does coredns/coredns go to `rancher/coredns-coredns` but metrics server doesn't go to `rancher/metrics-server-metrics-server`?

images-list
```diff
@@ -46,3 +46,6 @@ directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-ad
...
+k8s.gcr.io/pause rancher/pause 3.1
```
deniseschannon, line 51: This already exists in DockerHub and is multi-arch: 
https://hub.docker.com/repository/docker/rancher/pause/tags?page=1

images-list
```diff
@@ -46,3 +46,6 @@ directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-ad
 bats/bats rancher/bats-bats v1.1.0
 curlimages/curl rancher/curlimages-curl 7.70.0
 busybox rancher/busybox 1.31.1
+coredns/coredns rancher/coredns-coredns 1.6.9
```
deniseschannon, line 49: This version already exists in Dockerhub and is multi-arch: 

https://hub.docker.com/repository/docker/rancher/coredns-coredns/tags?page=1

deniseschannon:
In a side note, I added the automated publisher perms to all of these repos for whenever we want to have drone build/push these images.

images-list
brandond: I have many questions about the lack of conventions in the image mirror. 
* library/nginx goes to rancher/nginx but should have been rancher/library-nginx. 
* busybox is really library/busybox (no prefix = library/) but goes to just rancher/busybox.
* coredns/coredns went to rancher/coredns-coredns. We also have rancher/coredns which is our fips-140 build of coredns/coredns. I only know that because I asked.
* k3s uses rancher/metrics-server (has been for a while), even though it should have been rancher/metrics-server-metrics-server (as you point out). Our v0.3.6 tag is also a multiarch manifest list, while upstream v0.3.6 (not available on k8s.gcr.io) is amd64 only, although there are v0.3.6-arm64 and v0.3.6-arm tags available with the appropriate architectures. I've sidestepped this by going to v0.3.7 which is multiarch upstream. Since I'm going to have to change the tag in k3s anyway I might as well go to metrics-server-metrics-server as well.

I [opened an issue](https://github.com/rancher/rancher/issues/28595) to suggest that we move our mirrored stuff out to a dedicated org, and @craig was going to work on moving our hardened images to another org at some point as well. Right now everything's all mushed in together and it's challenging to pick apart because we've already got references to the current mishmash embedded in a bunch of supported releases.

images-list
brandond: Yes, but it was pushed manually by Darren a year ago. Should be automated.

images-list
brandond: Yes, but it was pushed manually by @cjellick. Should be automated.

brandond:
I just moved this back to draft to prevent merging until the mirror script is updated to support multiarch.

superseb:
What's the status of this PR?

brandond:
Will review soon. I've been focused on RKE2 GA tasks.

PR #31, opened 2020-09-09
Add Kiali version to 1.23.0

brendarearden:
Add Kiali version to 1.23.0

**Issue**
https://github.com/rancher/rancher/issues/28738


brendarearden:
I appended at the end instead of editing the original per the most recent instructions for updating image mirrors

cbron (approved):

PR #32, opened 2020-09-11
added images for logging operator 3.6.0

paynejacob:


cbron (changes requested):
Can you apply to bottom, its meant to be a running log of all new images added. Which kinda sucks cause this is much better to review.

cbron (approved):

brendarearden (approved):
🤞 looks good

PR #35, opened 2020-09-16
Add images for v1 rancher-monitoring and rancher-external-dns

PennyScissors:
Adding images used for security updates in v1 `rancher-monitoring` and `rancher-external-dns`

rancher-monitoring PR: https://github.com/rancher/system-charts/pull/334
rancher-external-dns PR: https://github.com/rancher/system-charts/pull/336

### DIFF
--- a/images-list
+++ b/images-list
@@ -1,20 +1,35 @@
-# This list should be sorted lexographically (alphabetically). Do not change existing entries. Comments lines may be started with # or //.
+# This list should be sorted in lexicographical (alphabetical) order. Do not change existing entries. Comments lines may be started with # or //.
 # This list was reset in order to prevent breakage when moving to multiarch mirroring. See the following PR for more information:
 # https://github.com/rancher/image-mirror/pull/27
 #
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
+bats/bats rancher/bats-bats v1.1.0
 coredns/coredns rancher/coredns-coredns 1.7.0
 coredns/coredns rancher/coredns-coredns 1.7.1
+curlimages/curl rancher/curlimages-curl 7.70.0
+directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-adapter-amd64 v0.6.0
+grafana/grafana rancher/grafana-grafana 7.1.5
 istio/coredns-plugin rancher/istio-coredns-plugin 0.2-istio-1.1
 istio/install-cni rancher/istio-install-cni 1.7.0
 istio/install-cni rancher/istio-install-cni 1.7.1
+istio/kubectl rancher/istio-kubectl 1.5.10
 istio/mixer rancher/istio-mixer 1.7.0
 istio/mixer rancher/istio-mixer 1.7.1
 istio/pilot rancher/istio-pilot 1.7.0
 istio/pilot rancher/istio-pilot 1.7.1
 istio/proxyv2 rancher/istio-proxyv2 1.7.0
 istio/proxyv2 rancher/istio-proxyv2 1.7.1
+jettech/kube-webhook-certgen rancher/jettech-kube-webhook-certgen v1.2.1
+jimmidyson/configmap-reload rancher/jimmidyson-configmap-reload v0.3.0
+kiwigrid/k8s-sidecar rancher/kiwigrid-k8s-sidecar 0.1.151
 library/busybox rancher/library-busybox 1.31.1
+library/nginx rancher/library-nginx 1.19.2-alpine
+openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.1.0
+prom/alertmanager rancher/prom-alertmanager v0.21.0
+prom/node-exporter rancher/prom-node-exporter v1.0.1
+prom/prometheus rancher/prom-prometheus v2.18.2
 quay.io/kiali/kiali rancher/kiali-kiali v1.22.1
 quay.io/kiali/kiali rancher/kiali-kiali v1.23.0
-
+squareup/ghostunnel rancher/squareup-ghostunnel v1.5.2
+thanosio/thanos rancher/thanosio-thanos v0.15.0
+us.gcr.io/k8s-artifacts-prod/external-dns/external-dns rancher/kubernetes-external-dns v0.7.3


images-list
```diff
@@ -3,4 +3,12 @@
 # https://github.com/rancher/image-mirror/pull/27
 #
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
+grafana/grafana rancher/grafana-grafana 7.1.5
+istio/kubectl rancher/istio-kubectl 1.5.10
 library/busybox rancher/library-busybox 1.31.1
+library/nginx rancher/nginx 1.19.2-alpine
```
brandond, line 9:
```suggestion
library/nginx rancher/library-nginx 1.19.2-alpine
```

images-list
```diff
@@ -3,4 +3,12 @@
...
+prom/alertmanager rancher/prom-alertmanager v0.21.0
+prom/node-exporter rancher/prom-node-exporter v1.0.1
+prom/prometheus rancher/prom-prometheus v2.21.0
+thanosio/thanos rancher/thanos v0.15.0
```
brandond, line 13:
```suggestion
thanosio/thanos rancher/thanos-thanos v0.15.0
```

brandond (changes requested):
Can we rename the destinations to keep with the <org>-<image> convention, or do you need them named as-is for historical reasons?

images-list
PennyScissors: @brandond Both nginx and thanos were in image-list with these names. I can rename them, but they would go into a new repository. Is that okay?

PennyScissors: @brandond I'll rename them, but new repositories are required since they were previously in image-list with these names.

cbron: We'll also need to update system-charts for new image name. @aiyengar2 this may affect monitoring v2 going forward.

brandond: @PennyScissors do you have access to create the new repos, or do we need to pull in someone else to do that? I don't personally have access to the rancher org.

images-list
```diff
@@ -3,4 +3,12 @@
 # https://github.com/rancher/image-mirror/pull/27
 #
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
+grafana/grafana rancher/grafana-grafana 7.1.5
+istio/kubectl rancher/istio-kubectl 1.5.10
 library/busybox rancher/library-busybox 1.31.1
+library/nginx rancher/nginx 1.19.2-alpine
+prom/alertmanager rancher/prom-alertmanager v0.21.0
+prom/node-exporter rancher/prom-node-exporter v1.0.1
+prom/prometheus rancher/prom-prometheus v2.21.0
```
aiyengar2, line 12: The current master chart uses version 2.18.2 https://github.com/prometheus-community/helm-charts/blob/4d780c4496a7df089dfde11c8731f26dc51c4a1f/charts/kube-prometheus-stack/values.yaml#L1617-L1618, which is also what we need for https://github.com/rancher/charts/pull/641.

aiyengar2 (changes requested):

images-list
cbron: Looks like needed in v1, so likely need both.

aiyengar2: @cbron the change isn't merged into v1 yet, so I left the same comment on that unmerged PR. https://github.com/rancher/system-charts/pull/334#discussion_r489709105

cbron: @aiyengar2 @PennyScissors The v1 monitoring and v2 monitoring tag versions can/should stay completely separate. They can use different tags and thats ok. What I meant above was that if we are changing the _name_ of the tag from one to another with this new framework that might affect both of them.

aiyengar2: @cbron not sure I agree that the tag versions should stay separate. 

It is true that Monitoring V1 / Monitoring V2 should have no connection to each other; however, both charts are still based off of the upstream chart at [`kube-prometheus-stack`](https://github.com/prometheus-community/helm-charts/tree/4d780c4496a7df089dfde11c8731f26dc51c4a1f/charts/kube-prometheus-stack), so to avoid any weird / extraneous issues that we would have to resolve (as opposed to upstream) it would be better if we can keep Monitoring V1 in sync with the images that are used by `kube-prometheus-stack`. 

If there's a specific requirement for Rancher for a higher version of Prometheus due to some specific feature request that only a higher version of Prometheus can support, it makes sense to have Monitoring V1 / V2 use different tag versions, but I don't think we've ever encountered such a use case.

PennyScissors: Ok, I'll downgrade to 2.18.2

images-list
aiyengar2: We don't use an nginx image afaik. @brendarearden are you using it for the broken magic links stuff?

brendarearden: @aiyengar2 I was able to prove out that it works to use the built in url workarounds, so I don't plan on using nginx.

PennyScissors: @aiyengar2 What does that mean exactly? Nginx is used here: https://github.com/rancher/system-charts/blob/7f97198bdd7cd39b87571a58fd5cde4f124bc49c/charts/rancher-monitoring/v0.1.2/values.yaml#L262

PennyScissors:
@brandond 
Renamed thanos and nginx in new commit

@cbron @aiyengar2 
I downgraded prometheus tag to v2.18.2, and also added gatekeepers image.
Gatekeeper also uses `rancher/kubectl:v1.18.6`, but I didn't add it here because it seems to be a custom image since it wasn't in the old image-list. Can any of you confirm?

brandond (approved):
LGTM on the image list.

I don't have any history on rancher/kubectl:v1.18.6 so I can't help with that :/

images-list
aiyengar2: Monitoring V1 uses it because it uses an authentication proxy in front of Grafana / Prometheus / Alertmanager in order to enforce Rancher ECM RBAC for Projects for users trying to access the UIs for those components.

Monitoring V2 does not utilize this since it doesn't need to support project-scoped monitoring

images-list
```diff
@@ -3,4 +3,13 @@
...
+grafana/grafana rancher/grafana-grafana 7.1.5
+istio/kubectl rancher/istio-kubectl 1.5.10
 library/busybox rancher/library-busybox 1.31.1
+library/nginx rancher/library-nginx 1.19.2-alpine
+openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.1.0
+prom/alertmanager rancher/prom-alertmanager v0.21.0
+prom/node-exporter rancher/prom-node-exporter v1.0.1
+prom/prometheus rancher/prom-prometheus v2.18.2
+thanosio/thanos rancher/thanosio-thanos v0.15.0
+us.gcr.io/k8s-artifacts-prod/external-dns/external-dns rancher/kubernetes-external-dns v0.7.3
```
aiyengar2, line 15: @PennyScissors  Do you think we can add in v2 rancher-monitoring images into this PR as well? It's gonna need the following additional images:
```
jimmidyson/configmap-reload rancher/jimmidyson-configmap-reload v0.3.0
kiwigrid/k8s-sidecar rancher/kiwigrid-k8s-sidecar 0.1.151
directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-adapter-amd64 v0.6.0
curlimages/curl rancher/curlimages-curl 7.70.0
busybox rancher/busybox 1.31.1
```

I can create it as a separate PR, but since it will be referring to a common subset of images (i.e. grafana, alertmanager, node-exporter, prometheus), it seems like it would be easier to keep them in one PR.

PennyScissors: Sure, I'll add them

aiyengar2:
> @brandond
> Renamed thanos and nginx in new commit
> 
> @cbron @aiyengar2
> I downgraded prometheus tag to v2.18.2, and also added a mirror for the gatekeeper's image `rancher/openpolicyagent-gatekeeper v3.1.0` that was in the previous image-list.
> Gatekeeper also uses `rancher/kubectl:v1.18.6`, but I didn't add it here because it seems to be a custom image since it wasn't in the old image-list. Can any of you confirm?

Seems like that is the case since we have a separate repo at https://github.com/rancher/kubectl which seems to maintain a tag-based drone pipeline.

images-list
cbron: We can stay on same version for now, that will let customers migrate easily between the two with same configuration. But if we are doing security updates in future we don't need to wait for other half to keep them in sync.

cbron (commented):
Are any of these pre-existing images in our dockerhub or are they all new ? I haven't checked.

aiyengar2 (dismissed):

images-list
```diff
@@ -3,4 +3,21 @@
...
+bats/bats rancher/bats-bats v1.1.0
+busybox rancher/busybox 1.31.1
+curlimages/curl rancher/curlimages-curl 7.70.0
+directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-adapter-amd64 v0.6.0
+grafana/grafana rancher/grafana-grafana 7.1.5
+istio/kubectl rancher/istio-kubectl 1.5.10
+jettech/kube-webhook-certgen rancher/jettech-kube-webhook-certgen v1.2.1
+jimmidyson/configmap-reload rancher/jimmidyson-configmap-reload v0.3.0
+kiwigrid/k8s-sidecar rancher/kiwigrid-k8s-sidecar 0.1.151
 library/busybox rancher/library-busybox 1.31.1
```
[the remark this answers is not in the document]
cbron: Actually @brandond is this still the case ? The readme at the top of file says alphabetical ?

PennyScissors:
> Are any of these pre-existing images in our dockerhub or are they all new ? I haven't checked.

Yes, gatekeeper's and I guess Arvind's are all pre-existing images.

images-list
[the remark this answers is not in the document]
cbron: nvm, its not sequential anymore

images-list
PennyScissors: @aiyengar2 Just noticed this image `busybox rancher/busybox 1.31.1` is already in images-list as `library/busybox rancher/library-busybox 1.31.1`. You might want to update the chart that uses it.

aiyengar2: Good call out, we do use it. Tracked in https://github.com/rancher/rancher/issues/28978

cbron (approved):
lgmt. None of these should be forked, or overwriting custom images. All were apparently mirrored previously.

PR #36, opened 2020-09-17
Add istio, kiali 2.5 images

brendarearden:

#### Types of Change ####

adding initial images ( mixture of new images for 1.7.1 and existing images for 1.7.0 ) 

#### Linked Issues ####

https://github.com/rancher/rancher/issues/28901

### DIFF
--- a/images-list
+++ b/images-list
@@ -1,6 +1,20 @@
-# This list should be sorted alphabetically. Do not change existing entries. Comments lines may be started with # or //.
+# This list should be sorted lexographically (alphabetically). Do not change existing entries. Comments lines may be started with # or //.
 # This list was reset in order to prevent breakage when moving to multiarch mirroring. See the following PR for more information:
 # https://github.com/rancher/image-mirror/pull/27
 #
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
+coredns/coredns rancher/coredns-coredns 1.7.0
+coredns/coredns rancher/coredns-coredns 1.7.1
+istio/coredns-plugin rancher/istio-coredns-plugin 0.2-istio-1.1
+istio/install-cni rancher/istio-install-cni 1.7.0
+istio/install-cni rancher/istio-install-cni 1.7.1
+istio/mixer rancher/istio-mixer 1.7.0
+istio/mixer rancher/istio-mixer 1.7.1
+istio/pilot rancher/istio-pilot 1.7.0
+istio/pilot rancher/istio-pilot 1.7.1
+istio/proxyv2 rancher/istio-proxyv2 1.7.0
+istio/proxyv2 rancher/istio-proxyv2 1.7.1
 library/busybox rancher/library-busybox 1.31.1
+quay.io/kiali/kiali rancher/kiali-kiali v1.22.1
+quay.io/kiali/kiali rancher/kiali-kiali v1.23.0
+


images-list
```diff
@@ -1,6 +1,19 @@
-# This list should be sorted alphabetically. Do not change existing entries. Comments lines may be started with # or //.
+# This list should be sorted lexographically. Do not change existing entries. Comments lines may be started with # or //.
 # This list was reset in order to prevent breakage when moving to multiarch mirroring. See the following PR for more information:
 # https://github.com/rancher/image-mirror/pull/27
 #
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
+coredns/coredns rancher/coredns-coredns 1.7.0
+coredns/coredns rancher/coredns-coredns 1.7.1
```
cbron, line 7: We could only do the version that will end up being live, but this is probably a good test case to check.

images-list
```diff
@@ -1,6 +1,19 @@
...
+istio/coredns-plugin rancher/istio-coredns-plugin 0.2-istio-1.1
```
cbron, line 8: Was thinking we got rid of coredns in 1.7, but maybe not. This looks to be only existing image in the PR.

images-list
```diff
@@ -1,6 +1,19 @@
...
+istio/install-cni rancher/istio-install-cni 1.7.1
+istio/mixer rancher/istio-mixer 1.7.0
+istio/mixer rancher/istio-mixer 1.7.1
+istio/pilot rancher/istio-pilot 1.7.0
+istio/pilot rancher/istio-pilot 1.7.1
+istio/proxyv2 rancher/istio-proxyv2 1.7.0
+istio/proxyv2 rancher/istio-proxyv2 1.7.1
 library/busybox rancher/library-busybox 1.31.1
+quay.io/kiali/kiali rancher/kiali-kiali v1.22.1
+quay.io/kiali/kiali rancher/kiali-kiali v1.23.0
```
cbron, line 19: nit: newline

cbron (commented):
Think this is ok. Would like another review.

images-list
brendarearden: From their annoucements, it seemed like they were going to phase out all `addonComponents` but coredns is not part of their [integrations](https://istio.io/latest/docs/ops/integrations/) list. The current demo profile for 1.7.1 still contains `istiocoredns` as an `addonComponents`

images-list
```diff
@@ -1,6 +1,20 @@
-# This list should be sorted alphabetically. Do not change existing entries. Comments lines may be started with # or //.
+# This list should be sorted lexographically. Do not change existing entries. Comments lines may be started with # or //.
```
brandond, line 1: Thanks for syncing this up with the PR template. I think I was waffling as to whether or not people would know what this meant ;)

brendarearden: I did have to google because I couldn't remember all the rules 😄

brandond (approved):

images-list
brandond: I think since we're including non-alphabetical characters (such as whitespace, punctuation, and digits) in the sort it's not technically in alphabetical order... so referring to it as lexographic order is technically correct - the best kind of correct!

cbron: I also had to google it 😆 . Since it's not a big deal if we accidentally fall into alphabetical instead of lexographical, let's  include alphabetical somewhere in here. Like: 
```suggestion
# This list should be sorted lexographically (alphabetically). Do not change existing entries. Comments lines may be started with # or //.
```

brendarearden: Alright, I went with caleb's approach so that you get a general idea of what to do, but could google for more accurate sorting.

cbron (approved):
lgtm

PR #37, opened 2020-09-17
Remove coredns from list

brendarearden:

- [ ] Change does not remove any existing Images or Tags
- [ ] New entries are added to the correct section of the list (sorted lexographically) 

#### Types of Change ####

Remove images that should not be in mirror images list

### DIFF
--- a/images-list
+++ b/images-list
@@ -5,7 +5,6 @@
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
 bats/bats rancher/bats-bats v1.1.0
 coredns/coredns rancher/coredns-coredns 1.7.0
-coredns/coredns rancher/coredns-coredns 1.7.1
 curlimages/curl rancher/curlimages-curl 7.70.0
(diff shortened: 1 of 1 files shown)


cbron (dismissed):
lgtm, but get brandon approval too just in case

images-list
```diff
@@ -3,8 +3,6 @@
 # https://github.com/rancher/image-mirror/pull/27
 #
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
-coredns/coredns rancher/coredns-coredns 1.7.0
```
brandond, line 6: We already pushed 1.7.0 so it should stay on the list just so that we know what we've done. 1.7.1 never got pushed since the source doesn't exist, so I'm OK with removing that one.

brandond (changes requested):

images-list
brendarearden: Updated to only remove 1.7.1 - makes sense to keep a record!

brandond (approved):

cbron (approved):

PR #39, opened 2020-09-19
Add gatekeeper v3.1.1

PennyScissors:
Add new image for gatekeeper v3.1.1: https://github.com/rancher/charts/pull/670

### DIFF
--- a/images-list
+++ b/images-list
@@ -3,9 +3,9 @@
 # https://github.com/rancher/image-mirror/pull/27
 #
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
-bats/bats rancher/bats-bats v1.1.0
-banzaicloud/logging-operator rancher/banzaicloud-logging-operator 3.6.0
 banzaicloud/fluentd rancher/banzaicloud-fluentd v1.11.2-alpine-2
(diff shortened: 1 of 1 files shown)


images-list
```diff
@@ -3,9 +3,9 @@
 # https://github.com/rancher/image-mirror/pull/27
 #
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
-bats/bats rancher/bats-bats v1.1.0
-banzaicloud/logging-operator rancher/banzaicloud-logging-operator 3.6.0
 banzaicloud/fluentd rancher/banzaicloud-fluentd v1.11.2-alpine-2
+banzaicloud/logging-operator rancher/banzaicloud-logging-operator 3.6.0
+bats/bats rancher/bats-bats v1.1.0
```
PennyScissors, line 8: @caleb @brandond These two images aren't gatekeeper's, but my script for sorting image-list caught them out of order. Should I remove this change or leave it in? I'm not sure about the consequences

cbron: Assuming its fine because the check would be on pull/push, but I'd like confirmation.

brandond: I'm going to add sort order and source existence checks to the PR CI script. Thanks for fixing this.

brandond (approved):

cbron (approved):

PR #40, opened 2020-09-24
Add Prometheus Operator v0.39.0 images

aiyengar2:

#### Types of Change ####

New images, but the repositories for both already exist:
- https://hub.docker.com/r/rancher/coreos-prometheus-operator
- https://hub.docker.com/r/rancher/coreos-prometheus-config-reloader

#### Linked Issues ####

Related Issue: https://github.com/rancher/rancher/issues/29149
Related PR: https://github.com/rancher/system-charts/pull/345

### DIFF
--- a/images-list
+++ b/images-list
@@ -31,6 +31,8 @@ openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.1.1
 prom/alertmanager rancher/prom-alertmanager v0.21.0
 prom/node-exporter rancher/prom-node-exporter v1.0.1
 prom/prometheus rancher/prom-prometheus v2.18.2
+quay.io/coreos/prometheus-operator rancher/coreos-prometheus-operator v0.39.0
+quay.io/coreos/prometheus-config-reloader rancher/coreos-prometheus-config-reloader v0.39.0
 quay.io/kiali/kiali rancher/kiali-kiali v1.22.1
 quay.io/kiali/kiali rancher/kiali-kiali v1.23.0
 squareup/ghostunnel rancher/squareup-ghostunnel v1.5.2


aiyengar2:
> New images, but the repositories for both already exist:
> - https://hub.docker.com/r/rancher/coreos-prometheus-operator
> - https://hub.docker.com/r/rancher/coreos-prometheus-config-reloader

Not sure if this requires any special consideration with the current state of the scripts; if we merge this PR, will it disturb the existing images in those repos? The older versions of these images are currently used by Monitoring V1 (and will continue to be used for Monitoring V2 since only V1 is upgrading to v0.39.0).

cc: @brandond @PennyScissors

PennyScissors:
> > New images, but the repositories for both already exist:
> > 
> > * https://hub.docker.com/r/rancher/coreos-prometheus-operator
> > * https://hub.docker.com/r/rancher/coreos-prometheus-config-reloader
> 
> Not sure if this requires any special consideration with the current state of the scripts; if we merge this PR, will it disturb the existing images in those repos? The older versions of these images are currently used by Monitoring V1 (and will continue to be used for Monitoring V2 since only V1 is upgrading to v0.39.0).
> 
> cc: @brandond @PennyScissors

I don't think it would cause any issues because v1 gatekeeper uses an older tag of the same image as v2 gatekeeper, so as long as they were pushed once it should fine. However, v2 gatekeeper uses a different repo now since before we made big changes in image-list so I'm not sure if any of those new changes could cause problems. @brandond might be able to confirm.

brandond (approved):
Looks to me like they're new tags for both, so LGTM

cbron (approved):

aiyengar2:
Thanks for the approvals! However, I'm holding off on merging this till https://github.com/rancher/system-charts/pull/345#issuecomment-698650663 is addressed.

aiyengar2:
https://github.com/rancher/rancher/issues/29149#issuecomment-699138151

PR #54, opened 2020-11-20
Add images for logging operator chart 3.8.0

nickgerace:

#### Types of Change ####

version bump

#### Linked Issues ####

Charts PR: https://github.com/rancher/charts/pull/857
Original Issue: https://github.com/rancher/rancher/issues/29892

#### Additional Notes ####

n/a

### DIFF
--- a/images-list
+++ b/images-list
@@ -5,9 +5,11 @@
 # LIST FORMAT: <SOURCE> <DESTINATION> <TAG>
 banzaicloud/fluentd rancher/banzaicloud-fluentd v1.11.2-alpine-2
 banzaicloud/fluentd rancher/banzaicloud-fluentd v1.11.4-alpine-1
+banzaicloud/fluentd rancher/banzaicloud-fluentd v1.11.5-alpine-1
 banzaicloud/logging-operator rancher/banzaicloud-logging-operator 3.6.0
 banzaicloud/logging-operator rancher/banzaicloud-logging-operator 3.7.0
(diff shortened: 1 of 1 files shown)


images-list
```diff
@@ -37,6 +41,7 @@ istio/proxyv2 rancher/istio-proxyv2 1.7.1
 istio/proxyv2 rancher/istio-proxyv2 1.7.3
 jettech/kube-webhook-certgen rancher/jettech-kube-webhook-certgen v1.2.1
 jettech/kube-webhook-certgen rancher/jettech-kube-webhook-certgen v1.5.0
+jimmidyson/configmap-reload rancher/jimmidyson-configmap-reload v0.2.2
```
cbron, line 44: Is this new ? Weird that we are adding an older version.

cbron: This already exists: https://hub.docker.com/r/rancher/jimmidyson-configmap-reload/tags?page=1&name=0.2

nickgerace: It does already exist, but I noticed it wasn't there. Should we add it regardless in the event that we need to pull and retag it again?

aiyengar2: Why is this being updated? I don't see the image change in https://github.com/rancher/charts/pull/857/files

nickgerace: My thought was that `images-list` was a source of truth in the event that we needed to repopulate the Rancher Docker registry. However, I spoke to Caleb and have since removed the image.

cbron (approved):

prachidamle (approved):

PR #55, opened 2020-11-20
Adds section to PR template to link dockerhub repos

cbron:
We should always be checking that the existing image does not exist first or we could overwrite it.
I do a lot of PR reviews on my phone, and without links that is very hard. This way the dev will do by hand first anyway.


cbron:
Handled by diff PR

PR #56, opened 2020-12-02
Add coredns 1.7.1 and 1.8.0

brandond:

- [X] Change does not remove any existing Images or Tags
- [X] New entries are in format `SOURCE DESTINATION TAG`
- [X] New entries are added to the correct section of the list (sorted lexicographically) 

#### Types of Change ####

* version bump

#### Linked Issues ####

k3s-io/k3s#2293

### DIFF
--- a/images-list
+++ b/images-list
@@ -12,6 +12,8 @@ banzaicloud/logging-operator rancher/banzaicloud-logging-operator 3.7.3
 banzaicloud/logging-operator rancher/banzaicloud-logging-operator 3.8.0
 bats/bats rancher/bats-bats v1.1.0
 coredns/coredns rancher/coredns-coredns 1.7.0
+coredns/coredns rancher/coredns-coredns 1.7.1
+coredns/coredns rancher/coredns-coredns 1.8.0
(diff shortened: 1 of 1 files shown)


cbron (approved):
confirmed these aren't pre-existing

images-list
```diff
@@ -12,6 +12,8 @@ banzaicloud/logging-operator rancher/banzaicloud-logging-operator 3.7.3
 banzaicloud/logging-operator rancher/banzaicloud-logging-operator 3.8.0
 bats/bats rancher/bats-bats v1.1.0
 coredns/coredns rancher/coredns-coredns 1.7.0
+coredns/coredns rancher/coredns-coredns 1.7.1
+coredns/coredns rancher/coredns-coredns 1.8.0
```
prachidamle, line 16: I have a question here - why do we maintain multiple versions of the same image?

brandond: Because different versions of k3s/rke2/rke use different versions of the image, and all versions need to be present in our mirror.

prachidamle (approved):

PR #47, opened 2020-11-06
Update images for Monitoring rebase to 11.0.0

aiyengar2:
These image updates are required for https://github.com/rancher/charts/pull/836.

#### Pull Request Checklist ####

- [x] Change does not remove any existing Images or Tags
- [x] New entries are in format `SOURCE DESTINATION TAG`
- [x] New entries are added to the correct section of the list (sorted lexographically) 
- [x] Changes to scripting or CI config have been tested to the best of your ability

#### Types of Change ####

New image tags for:
- curlimages/curl rancher/curlimages-curl 7.73.0
- grafana/grafana rancher/grafana-grafana 7.2.1
- jettech/kube-webhook-certgen rancher/jettech-kube-webhook-certgen v1.5.0
- jimmidyson/configmap-reload rancher/jimmidyson-configmap-reload v0.4.0
- kiwigrid/k8s-sidecar rancher/kiwigrid-k8s-sidecar 1.1.0

New images that require new repositories:
- grafana/grafana-image-renderer rancher/grafana-grafana-image-renderer 2.0.0
    - Created a ticket to create the new repository

New images that have existing Rancher repositories but have been added to the scripts for the first time:
- quay.io/prometheus-operator/prometheus-config-reloader rancher/coreos-prometheus-config-reloader v0.43.0
    - https://hub.docker.com/r/rancher/coreos-prometheus-config-reloader/tags
- quay.io/prometheus-operator/prometheus-operator rancher/coreos-prometheus-operator v0.43.0
    - https://hub.docker.com/r/rancher/coreos-prometheus-operator/tags
- quay.io/prometheus/prometheus rancher/prom-prometheus v2.22.0
    - https://hub.docker.com/r/rancher/prom-prometheus/tags

#### Linked Issues ####

PR: https://github.com/rancher/charts/pull/836

Related Issue: https://github.com/rancher/rancher/issues/29871

PR #76, opened 2021-02-18
Add 'mirrored-' to mirrored charts for 2.5.6

thedadams:

#### Types of Change ####
New repos for the `mirrored-` images.

#### Linked Issues ####
https://github.com/rancher/rancher/issues/30888
https://github.com/rancherlabs/eio/issues/369

#### Related PRs ####
https://github.com/rancher/charts/pull/1003


images-list
```diff
@@ -52,48 +63,70 @@ istio/pilot rancher/istio-pilot 1.7.3
...
 istio/proxyv2 rancher/istio-proxyv2 1.8.3
+istio/proxyv2 rancher/mirrored-istio-proxyv2 1.8.3
 jaegertracing/all-in-one rancher/jaegertracing-all-in-one 1.20.0
+jaegertracing/all-in-one rancher/mirrored-jaegertracing-all-in-one 1.20.0
 jettech/kube-webhook-certgen rancher/jettech-kube-webhook-certgen v1.2.1
 jettech/kube-webhook-certgen rancher/jettech-kube-webhook-certgen v1.5.0
+jettech/kube-webhook-certgen rancher/mirrored-jettech-kube-webhook-certgen v1.2.1
 jimmidyson/configmap-reload rancher/jimmidyson-configmap-reload v0.3.0
 jimmidyson/configmap-reload rancher/jimmidyson-configmap-reload v0.4.0
+jimmidyson/configmap-reload rancher/mirrored-jimmidyson-configmap-reload v0.3.0
```
cbron, line 81: We need both 0.4.0 and 0.3.0

images-list
```diff
@@ -120,20 +153,29 @@ quay.io/cilium/operator-aws rancher/mirrored-cilium-operator-aws v1.9.4
...
 quay.io/coreos/prometheus-operator rancher/coreos-prometheus-operator v0.39.0
+quay.io/coreos/prometheus-operator rancher/mirrored-coreos-prometheus-operator v0.38.1
 quay.io/kiali/kiali rancher/kiali-kiali v1.22.1
 quay.io/kiali/kiali rancher/kiali-kiali v1.23.0
 quay.io/kiali/kiali rancher/kiali-kiali v1.24.0
 quay.io/kiali/kiali rancher/kiali-kiali v1.28.0
 quay.io/kiali/kiali rancher/kiali-kiali v1.29.0
+quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.29.0
 quay.io/prometheus-operator/prometheus-config-reloader rancher/coreos-prometheus-config-reloader v0.43.0
+quay.io/prometheus-operator/prometheus-config-reloader rancher/mirrored-coreos-prometheus-config-reloader v0.43.0
```
cbron, line 168: Did we find out why its coming from 2 different origins ? Technically we shouldn't be doing this. Was it for ease of chart change ?

images-list
```diff
@@ -120,20 +153,29 @@ quay.io/cilium/operator-aws rancher/mirrored-cilium-operator-aws v1.9.4
...
+quay.io/prometheus/prometheus rancher/mirrored-prom-prometheus v2.18.2
 quay.io/prometheus/prometheus rancher/prom-prometheus v2.19.2
 quay.io/prometheus/prometheus rancher/prom-prometheus v2.22.0
 rpardini/docker-registry-proxy rancher/rpardini-docker-registry-proxy 0.6.1
+sonobuoy/sonobuoy rancher/mirrored-sonobuoy-sonobuoy v0.16.3
 sonobuoy/sonobuoy rancher/sonobuoy-sonobuoy v0.19.0
+squareup/ghostunnel rancher/mirrored-squareup-ghostunnel v1.5.2
 squareup/ghostunnel rancher/squareup-ghostunnel v1.5.2
 thanosio/thanos rancher/thanosio-thanos v0.15.0
 us.gcr.io/k8s-artifacts-prod/external-dns/external-dns rancher/kubernetes-external-dns v0.7.3
```
cbron, line 181: should leave blankline at bottom

cbron (changes requested):

images-list
thedadams: I went through the versions in rancher/charts and didn't seen v0.4.0 for this image. Do I need to do this for all versions, not just the ones in rancher/charts?

images-list
thedadams: No, I asked Arvind about something else, but forgot to ask about this.

images-list
thedadams: Will do.

images-list
cbron: https://github.com/rancher/charts/pull/1003/files#diff-5bb2300e45effc1f02a9d2fd7e3ca39d7efc30c030d641a30164056cc332f367R67
We should scan that other PR for all these images and see if they are used in multiple places

images-list
cbron: @aiyengar2 you remember why we did this ?

images-list
thedadams: Sorry. I did that, but I missed one. I went through it again and this is the only one I missed.

images-list
thedadams: There were three that I found that were like this. I removed the ones that were not explicitly mentioned in https://github.com/rancher/charts/pull/1003. Those were:
prom/prometheus rancher/mirrored-prom-prometheus v2.18.2
quay.io/prometheus-operator/prometheus-operator rancher/mirrored-coreos-prometheus-operator v0.43.0
quay.io/prometheus-operator/prometheus-config-reloader rancher/mirrored-coreos-prometheus-config-reloader v0.43.0

I can re-add them if necessary.

cbron (approved):

images-list
cbron: The first one is used in system-charts, that will be updated later. The other 2 come from this PR: https://github.com/rancher/image-mirror/pull/47/files
Looks like that isn't in yet, so I don't think we need to mirror those. When we do we should try to put them in the right org.

cbron: Confirmed that we don't need those yet. 0.39.x is used in system-charts right now but we aren't mirroring those for 2.5.6. And the 0.43 ones aren't migrated to yet.

PR #88, opened 2021-04-08
Add kiali v1.32.0

brendarearden:

#### Types of Change ####

Version Bump

#### Linked Issues ####
https://github.com/rancher/rancher/issues/31979

### DIFF
--- a/images-list
+++ b/images-list
@@ -235,6 +235,7 @@ quay.io/kiali/kiali rancher/kiali-kiali v1.24.0
 quay.io/kiali/kiali rancher/kiali-kiali v1.28.0
 quay.io/kiali/kiali rancher/kiali-kiali v1.29.0
 quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.29.0
+quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.32.0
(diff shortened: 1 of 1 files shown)


images-list
```diff
@@ -235,6 +235,7 @@ quay.io/kiali/kiali rancher/kiali-kiali v1.24.0
 quay.io/kiali/kiali rancher/kiali-kiali v1.28.0
 quay.io/kiali/kiali rancher/kiali-kiali v1.29.0
 quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.29.0
+quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.32.0
```
cbron, line 238: wasn't there a tracing change ?

brendarearden: The latest version of jaeger tracing is 1.22 - should we go ahead and update to that as well?

cbron: We don't have to, your call, I'm not sure whats been released or how that version syncs with the new istio versions.

brendarearden: Since the jaeger sample that is being used with the latest version of istio is still v1.20 and there isn't a compatibility table else where, we should stick with 1.20 for now. https://raw.githubusercontent.com/istio/istio/release-1.9/samples/addons/jaeger.yaml

cbron (approved):

PR #89, opened 2021-04-08
add entry for CoreDNS 1.8.3

briandowns:

- [ x ] Change does not remove any existing Images or Tags in index-list.txt
- [ x ] Change does not remove / overwrite exiting Images or Tags in Rancher DockerHub
- [ x ] New entries are in format `SOURCE DESTINATION TAG`
- [ x ] New entries are added to the correct section of the list (sorted lexographically)
- [ x ] New entries have a repo created in Rancher Dockerhub (where the image will be mirrored to)

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/3001

### DIFF
--- a/images-list
+++ b/images-list
@@ -23,6 +23,7 @@ coredns/coredns rancher/coredns-coredns 1.7.0
 coredns/coredns rancher/coredns-coredns 1.7.1
 coredns/coredns rancher/coredns-coredns 1.8.0
 coredns/coredns rancher/coredns-coredns 1.8.1
+coredns/coredns rancher/coredns-coredns 1.8.3
 coredns/coredns rancher/mirrored-coredns-coredns 1.6.2
 coredns/coredns rancher/mirrored-coredns-coredns 1.6.5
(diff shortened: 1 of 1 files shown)


brandond (approved):

images-list
```diff
@@ -23,6 +23,7 @@ coredns/coredns rancher/coredns-coredns 1.7.0
 coredns/coredns rancher/coredns-coredns 1.7.1
 coredns/coredns rancher/coredns-coredns 1.8.0
 coredns/coredns rancher/coredns-coredns 1.8.1
+coredns/coredns rancher/coredns-coredns 1.8.3
```
cbron, line 26: Is it possible to move code to rancher/mirrored-coredns-coredns ? If not we can move forward as-is. But want to do that soon.

cbron (approved):

images-list
briandowns: I'm fine to do that. This is due for a dev complete of tomorrow. Can we do this after the next k3s release?

PR #96, opened 2021-04-28
add the missing mirrored images for monitoring v1 chart version 0.2.2

jiaqiluo:

#### Types of Change ####

New image


images-list
```diff
@@ -91,6 +91,7 @@ istio/install-cni rancher/mirrored-istio-install-cni 1.9.2
 istio/install-cni rancher/mirrored-istio-install-cni 1.9.3
 istio/kubectl rancher/istio-kubectl 1.5.10
 istio/kubectl rancher/mirrored-istio-kubectl 1.5.9
+istio/kubectl rancher/mirrored-istio-kubectl 1.5.10
```
cbron, line 94: Whats this for again ? That image is very old.

images-list
```diff
@@ -208,6 +209,7 @@ prom/node-exporter rancher/mirrored-prom-node-exporter v1.0.1
 prom/node-exporter rancher/prom-node-exporter v1.0.1
 prom/prometheus rancher/prom-prometheus v2.18.2
 prom/prometheus rancher/mirrored-prom-prometheus v2.12.0
+prom/prometheus rancher/mirrored-prom-prometheus v2.18.2
```
cbron, line 212: This exists in lower section as it moved. @aiyengar2 this is a problem we will run into since we didn't change the image name.

cbron (changes requested):

images-list
jiaqiluo: It appears on the images.txt of the latest rancher v2.5.8-rc12
ttps://github.com/rancher/rancher/releases/tag/v2.5.8-rc12 

and it is used here: https://github.com/rancher/system-charts/blob/7e6a27f0e377beccea51935b5af49cfc77e243b0/charts/rancher-monitoring/v0.2.2/values.yaml#L394-L395

images-list
jiaqiluo: Good catch!  
I removed the duplicated line.

images-list
cbron: Ok, we can do this for now. But @aiyengar2 I think we do need long term solution here. Can we use rancher/kubectl ?

aiyengar2: Yup, let's swap it. This just gets used in to run `command: ["kubectl",  "apply", "-f", "/etc/operator-init/."]`:

https://github.com/rancher/system-charts/blob/7e6a27f0e377beccea51935b5af49cfc77e243b0/charts/rancher-monitoring/v0.2.2/charts/operator-init/templates/job-install-crds.yaml#L33

aiyengar2 (changes requested):

images-list
cbron: @jiaqiluo there have been changes in this repo, you'll need to rebase

jiaqiluo: @aiyengar2 @cbron  I am going to update the monitoring chart to use `rancher/kubectl:v1.20.2`. 

Should the PR for the monitoring chart go `dev-v2.5-source` or `dev-v2.6-source`?

cbron: I'd update both so they stay in version line. Some point soon 2.5-source and 2.6-source will both have full folder on them. 
Can you also rebase this PR

jiaqiluo:
@cbron @aiyengar2  
This PR is ready for final review and merging. 

Monitoring v0.2.3 will use the mirrored images.

images-list
```diff
@@ -330,5 +332,6 @@ sonobuoy/sonobuoy rancher/mirrored-sonobuoy-sonobuoy v0.16.3
 sonobuoy/sonobuoy rancher/sonobuoy-sonobuoy v0.19.0
 squareup/ghostunnel rancher/mirrored-squareup-ghostunnel v1.5.2
 squareup/ghostunnel rancher/squareup-ghostunnel v1.5.2
+thanosio/thanos rancher/mirrored-thanosio-thanos v0.15.0
```
aiyengar2, line 335: Why are we using thanos? Monitoring V1 doesn't have support for it

images-list
```diff
@@ -280,8 +280,10 @@ quay.io/cilium/startup-script rancher/mirrored-cilium-startup-script 62bfbe88c17
 quay.io/coreos/kube-state-metrics rancher/mirrored-coreos-kube-state-metrics v1.9.7
 quay.io/coreos/prometheus-config-reloader rancher/coreos-prometheus-config-reloader v0.39.0
 quay.io/coreos/prometheus-config-reloader rancher/mirrored-coreos-prometheus-config-reloader v0.38.1
+quay.io/coreos/prometheus-config-reloader rancher/mirrored-coreos-prometheus-config-reloader v0.39.0
 quay.io/coreos/prometheus-operator rancher/coreos-prometheus-operator v0.39.0
 quay.io/coreos/prometheus-operator rancher/mirrored-coreos-prometheus-operator v0.38.1
+quay.io/coreos/prometheus-operator rancher/mirrored-coreos-prometheus-operator v0.39.0
```
aiyengar2, line 286: Hmm, seems like we are using the older quay.io repositories here, but I checked the new ones at `quay.io/prometheus-operator/prometheus-config-reloader` and `quay.io/prometheus-operator/prometheus-operator` and they don't have the v0.39.0 image tags.

Looks good to me, but just wanted to leave this as a comment so, if we update Monitoring V1, we make sure we migrate to the new repositories next time.

aiyengar2 (commented):
Unsure about Thanos image, but otherwise LGTM

images-list
aiyengar2: https://github.com/rancher/system-charts/blob/4960259f853d6dc10fe8fd346cd38423daf42626/charts/rancher-monitoring/v0.2.2/charts/prometheus/templates/prometheus.yaml#L98-L103 apparently we do support Thanos in Monitoring V1, although it's probably only an option configurable as part of the overrides in Monitoring V1 today.

aiyengar2 (approved):
LGTM

PR #98, opened 2021-04-29
Update sort validator to use version sort and sort current images

PennyScissors:


.drone.yml
```diff
@@ -7,30 +7,29 @@ platform:
...
-  - cat images-list | dapper
-  settings:
-    custom_dns: 1.1.1.1
-  when:
-    event:
-    - push
+  - name: validate-list
+    image: library/busybox:1.33.0
+    commands:
+      - grep -vE '^\s*(#|//)' images-list | sort -Vc
```
PennyScissors, line 13: The only significant change here is sort's flags, the rest is just lint/format.

.drone.yml
```diff
@@ -7,30 +7,29 @@ platform:
...
-  - cat images-list | dapper
-  settings:
-    custom_dns: 1.1.1.1
-  when:
-    event:
-    - push
+  - name: validate-list
+    image: library/ubuntu:21.04
+    commands:
+      - grep -vE '^\s*(#|//)' images-list | sort -Vc
```
PennyScissors, line 13: The only significant changes are:
1. Added the flag `-V` to sort so that it does version sorting
2. Changed the image to ubuntu. The reason for this is that busybox and alpine have a different version of sort that doesn't sort the images in the same way and the validation fails. Ubuntu's sort seems to match my mac's sort and passes validation without problems, so it's a better choice considering most of us work in Macs or a debian based OS.

PennyScissors:
@cbron The last commit adds a script to sort the images. Basically, you add all the images you want to add at the end of the images-list file, then run the script to generate a new images-list file that has all the images sorted in the correct order. I'm not sure if this is something that you would like to keep here, or if it should be implemented in the CI instead, but I figured I would commit it so that you check it out and decide if we should keep it.

.drone.yml
brandond: What's with all the whitespace changes?

sort-images-list.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Instructions: Add the new images at the end of the images-list file, then run this
+# script to generate a new images-list file with all images sorted in the correct order
+
+set -e
+
+HEADER="# This list should be sorted in lexicographical (alphabetical) order. Do not change existing entries. Comments lines may be started with # or //.\n# This list was reset in order to prevent breakage when moving to multiarch mirroring. See the following PR for more information:\n# https://github.com/rancher/image-mirror/pull/27\n#\n# LIST FORMAT: <SOURCE> <DESTINATION> <TAG>"
```
brandond, line 7: The header says alphabetical but a version sort isn't technically the same thing as an alphabetical sort. We should decide which one we want and make sure that the documentation and tooling agree.

brandond (changes requested):

sort-images-list.sh
brandond: I also don't like that this strips out all comments other than the header; the intent was to allow for comments anywhere in the file as long as they start with whitespace followed by # or //

PennyScissors: In that case I can remove the script. I wasn't aware we wanted to allow comments in this file since we don't have any at the moment and I'm not sure in what situation leaving a comment in this file would be helpful. 

As for the sorting, I think version sort would be a better fit because using the `-d` flag will only look at the first digit of a number and result in some versions being out of place.

For example (see v3.4.3-rancher1):
sort with `-d`
```
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.3.15-rancher1
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.4.13-rancher1
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.4.14-rancher1
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.4.15-rancher1
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.4.3-rancher1
```

sort with `-V`
```
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.3.15-rancher1
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.4.3-rancher1
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.4.13-rancher1
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.4.14-rancher1
rancher/coreos-etcd rancher/mirrored-coreos-etcd v3.4.15-rancher1
```

@cbron Any thoughts?

brandond: I'm OK with the CI check using a version sort, and expecting folks creating PRs to do the same. I think the helper script is great as a tool for folks to check what they're adding - maybe we just don't set the expectation that folks blindly run it and commit the output?

PennyScissors: I agree. My train of thought is that since we now have validation for the sorting (Should never get merged our of order again),  and we don't allow removing or overwriting images, we can expect all PRs to only add images, and it would be easy to spot if something is wrong after running the script.

.drone.yml
PennyScissors: The whitespace is just format/lint adjusting the indentation.

brandond (approved):

cbron (approved):

PR #105, opened 2021-05-10
correct source repo from gcr.io/ to k8s.gcr.io/ for kube-dns

kinarashah:


StrongMonkey (approved):

images-list
```diff
@@ -56,19 +56,15 @@ gcr.io/cloud-provider-vsphere/csi/release/syncer rancher/mirrored-cloud-provider
...
 gcr.io/google_containers/k8s-dns-dnsmasq-nanny rancher/mirrored-k8s-dns-dnsmasq-nanny 1.15.10
-gcr.io/google_containers/k8s-dns-dnsmasq-nanny rancher/mirrored-k8s-dns-dnsmasq-nanny 1.17.3
 gcr.io/google_containers/k8s-dns-kube-dns rancher/mirrored-k8s-dns-kube-dns 1.15.0
 gcr.io/google_containers/k8s-dns-kube-dns rancher/mirrored-k8s-dns-kube-dns 1.15.2
 gcr.io/google_containers/k8s-dns-kube-dns rancher/mirrored-k8s-dns-kube-dns 1.15.10
-gcr.io/google_containers/k8s-dns-kube-dns rancher/mirrored-k8s-dns-kube-dns 1.17.3
 gcr.io/google_containers/k8s-dns-sidecar rancher/mirrored-k8s-dns-sidecar 1.15.0
 gcr.io/google_containers/k8s-dns-sidecar rancher/mirrored-k8s-dns-sidecar 1.15.2
 gcr.io/google_containers/k8s-dns-sidecar rancher/mirrored-k8s-dns-sidecar 1.15.10
-gcr.io/google_containers/k8s-dns-sidecar rancher/mirrored-k8s-dns-sidecar 1.17.3
```
cbron, line 71: So we are actively overwriting the existing images ? Are these used in prod right now ? And are the images different or same. 
Typically we never delete so as to keep a history but I suppose this is special case.

cbron: I suppose if we didn't it would be ping/ponging back and forth, and that would be bad.

kinarashah: It isn't in prod, I added this recently for 1.17.3 (new version) - but seems like upstream images no longer exist `gcr.io` so the images weren't pushed at all.  Docker pull failed in the build logs but I missed them because the build passed, so replacing them now. 
TLDR - this seems like a special case, and I'd make sure to pull from source repo now onwards to make sure we don't have to remove anything like this

PR #146, opened 2021-08-03
add new image rancher/mirrored-library-traefik

jiaqiluo:

#### Types of Change ####

New image

#### Linked Issues ####

https://github.com/k3s-io/k3s/pull/3749

### DIFF
--- a/images-list
+++ b/images-list
@@ -214,6 +214,7 @@ library/nginx rancher/mirrored-library-nginx 1.21.1-alpine
 library/traefik rancher/library-traefik 2.4.2
 library/traefik rancher/library-traefik 2.4.8
 library/traefik rancher/library-traefik 2.4.9
+library/traefik rancher/mirrored-library-traefik 2.4.8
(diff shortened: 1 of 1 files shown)


images-list
```diff
@@ -214,6 +214,8 @@ library/nginx rancher/mirrored-library-nginx 1.21.1-alpine
 library/traefik rancher/library-traefik 2.4.2
 library/traefik rancher/library-traefik 2.4.8
 library/traefik rancher/library-traefik 2.4.9
+library/traefik rancher/mirrored-library-traefik 2.4.8
+library/traefik rancher/mirrored-library-traefik 2.4.9
```
cbron, line 218: Why both ? I don't see 2.4.9 in the other PR.

jiaqiluo: I tried to match the version between the old `rancher/library-traefik` and new `rancher/mirrored-library-traefik`, so I add all versions from 2.4.8 in case we will need 2.4.9 in the near future. 

But I can remove it for now and add it when it is needed.

cbron (approved):

PR #156, opened 2021-09-02
Add mirrored image for gatekeeper 3.6.0 and gatekeeper-crds 3.6.0

rayandas:

#### Types of Change ####

New image, version bump.

#### Linked Issues ####

https://github.com/rancher/rancher/issues/34593

### DIFF
--- a/images-list
+++ b/images-list
@@ -311,10 +311,12 @@ openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.3.0
 openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.4.0
 openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.5.1
 openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.5.2
+openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.6.0
 openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.1.0
 openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.1.1
 openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.2.1
 openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.3.0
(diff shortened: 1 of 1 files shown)


images-list
```diff
@@ -311,10 +311,12 @@ openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.3.0
 openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.4.0
 openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.5.1
 openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.5.2
+openpolicyagent/gatekeeper rancher/mirrored-openpolicyagent-gatekeeper v3.6.0
 openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.1.0
 openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.1.1
 openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.2.1
 openpolicyagent/gatekeeper rancher/openpolicyagent-gatekeeper v3.3.0
+openpolicyagent/gatekeeper-crds rancher/mirrored-openpolicyagent-gatekeeper-crds v3.6.0
```
cbron, line 319: @PennyScissors do we need the crds here ? we haven't mirrored that before.

rayandas: Gatekeeper upstream started adding crd image in `values.yaml` from `v3.6.0` you can find it [here](https://github.com/open-policy-agent/gatekeeper/blob/19ee221352a4d34ea659512e5ab376bae5fc17b8/charts/gatekeeper/values.yaml#L28).
In earlier versions there was only one image. For example if you check it for `v3.5.1` there was only the gatekeeper image. It's the [permalink](https://github.com/open-policy-agent/gatekeeper/blob/c8621550a3cc38d11f08b33a5c2b74a9ad328b1f/charts/gatekeeper/values.yaml#L24) for reference.

rayandas: Also a new repo has been created for crds. Here is the issue. https://github.com/rancherlabs/eio/issues/563

PennyScissors: @cbron That's correct, upstream added this new image in v3.6.0 so we need to mirror it.
image

cbron (approved):
Repo exists so we are ok

PR #161, opened 2021-09-09
Add istio-1.11.2

doflamingo721:

#### Types of Change ####
version bump

#### Linked Issues ####
https://github.com/rancher/rancher/issues/33683

#### Additional Notes ####
N/A


brendarearden (approved):
LGTM! I think previously you only had 1 the cni image with a version bump, but when upgrading istio, you have to bump all of the related images to the same version - so this is good!

doflamingo721:
> LGTM! I think previously you only had 1 the cni image with a version bump, but when upgrading istio, you have to bump all of the related images to the same version - so this is good!

Yes, that's correct. Before I was upgrading the `install-cni` image for resolving CVEs but later started on updating istio chart so added the other images as well.

kinarashah (approved):

brendarearden:
@kinarashah I think that means you can go ahead and merge this one in

PR #173, opened 2021-10-14
changing to mirrored

luthermonson:
per @cbron switching to mirrored- prefix for flannel cni plugin


kinarashah (commented):
Mirrored images are what we usually use, but would like to know what changed after https://github.com/rancher/image-mirror/pull/171#pullrequestreview-776802135

images-list
```diff
@@ -46,7 +46,7 @@ directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-ad
 directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-adapter-amd64 v0.7.0
 directxman12/k8s-prometheus-adapter-amd64 rancher/mirrored-directxman12-k8s-prometheus-adapter-amd64 v0.7.0
 directxman12/k8s-prometheus-adapter-amd64 rancher/mirrored-directxman12-k8s-prometheus-adapter-amd64 v0.8.3
-flannelcni/flannel-cni-plugin rancher/flannel-cni-plugin v1.2
+flannelcni/flannel-cni-plugin rancher/mirrored-flannel-cni-plugin v1.2
```
cbron, line 49: Missing the org
```suggestion
flannelcni/flannel-cni-plugin rancher/mirrored-flannelcni-flannel-cni-plugin v1.2
```

luthermonson:
The official direction is to do mirrored and this will be permanent and used by the community. Would rather do it the right way as we already do mirrored-flannel

cbron (approved):

images-list
```diff
@@ -46,7 +46,7 @@ directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-ad
 directxman12/k8s-prometheus-adapter-amd64 rancher/directxman12-k8s-prometheus-adapter-amd64 v0.7.0
 directxman12/k8s-prometheus-adapter-amd64 rancher/mirrored-directxman12-k8s-prometheus-adapter-amd64 v0.7.0
 directxman12/k8s-prometheus-adapter-amd64 rancher/mirrored-directxman12-k8s-prometheus-adapter-amd64 v0.8.3
-flannelcni/flannel-cni-plugin rancher/flannel-cni-plugin v1.2
```
cbron, line 49: This repo is gone now, so we are removing so it doesn't fail to sync every time.

PR #175, opened 2021-10-19
Create mirror images for s390x Arch

ivan-claire:

- [ ] Change does not remove any existing Images or Tags in the images-list file
- [ ] Change does not remove / overwrite exiting Images or Tags in Rancher DockerHub

#### Additional Notes ####
VERIFIED that images that do not contain s390x archs are not affected. [http://35.180.252.184/ivan-claire/image-mirror/8](in this test drone CI)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -12,7 +12,7 @@ steps:
     commands:
       - grep -vE '^\s*(#|//)' images-list | sort -Vc
   - name: mirror-images
-    image: rancher/dapper:v0.5.3
+    image: rancher/dapper:v0.5.7
     volumes:
       - name: docker
         path: /var/run/docker.sock
--- a/image-mirror.sh
+++ b/image-mirror.sh
@@ -3,7 +3,7 @@
 set -uo pipefail
 export DOCKER_CLI_EXPERIMENTAL="enabled"
 
-ARCH_LIST="amd64 arm64 arm"
+ARCH_LIST="amd64 arm64 arm s390x"
 
 function copy_if_changed {
   SOURCE_REF="${1}"


.drone.yml
```diff
@@ -12,7 +12,7 @@ steps:
     commands:
       - grep -vE '^\s*(#|//)' images-list | sort -Vc
   - name: mirror-images
-    image: rancher/dapper:v0.5.3
+    image: rancher/dapper:v0.5.7
```
dweomer, line 15: This looks fine to me but IIRC anything past dapper 0.5.0 segfaults on arm (i am not super familiar with this repo and what it does, actually)

cjellick:
@ivan-claire What mirrored image(s) do you need s390x arch for? ...I ask because offhand I thought that rke2's images didn't really use mirrored images. May be some I'm forgetting.

Note for reviewers: this is not a blocker to merging if you are good with the change. I'm just curious

.drone.yml
cbron: That shouldn't actually matter, we pull/push the manifest based on the OS's specified in the line changed below, not what OS you are currently running on.

cbron: I am curious though why we upgraded, if there was something new in dapper we needed to pull in, or if this is just a generic upgrade ?

.drone.yml
```diff
@@ -12,7 +12,7 @@ steps:
     commands:
```
cbron, line 12: Please review/check boxes in the description.

image-mirror.sh
```diff
@@ -3,7 +3,7 @@
 set -uo pipefail
 export DOCKER_CLI_EXPERIMENTAL="enabled"
 
-ARCH_LIST="amd64 arm64 arm"
+ARCH_LIST="amd64 arm64 arm s390x"
```
cbron, line 6: Is there a image we are targeting with this ? How would we know if it is working.

.drone.yml
ivan-claire: we need to upgrade to 0.5.7 because 0.5.7 has an s390x arch image published to dockerhub while 0.5.3 doesn't.
Without it, the s390x images won't be built properly.

ivan-claire: @cjellick  there are 2 mirrored images in the list that are being used by rke2:
mirrored-pause  and mirrored-ingress-nginx-kube-webhook-certgen

image-mirror.sh
manuelbuil: For example, it'll be useful for https://hub.docker.com/r/flannelcni/flannel-cni-plugin/tags. Currently, the s390x arch is not being mirrored https://hub.docker.com/r/rancher/mirrored-flannelcni-flannel-cni-plugin/tags

cbron (approved):
lgtm, someone will need to watch the build and ensure no errors exist, and that images are mirrored correctly.

PR #183, opened 2021-10-29
Add new flannel image

manuelbuil:
Signed-off-by: Manuel Buil <mbuil@suse.com>


kinarashah (approved):
Removed v1.0 because it's no longer valid image. flannel plugin versions would now follow vX.X.X format

PR #186, opened 2021-11-09
Don't update the manifest list if the list of images it refers to hasn't changed

brandond:

#### Types of Change ####

script update

#### Linked Issues ####

@Oats87 

#### Additional Notes ####

Chris noted that the script was causing the "last changed" time on the manifest lists to get updated every time the script runs. This is because we were rewriting the manifest list every time the script is run, even if we didn't sync any new images.

This change compares the digests in the new manifest list to those in the current manifest list (if any) and only pushes the manifest list if they do not match.

### DIFF
--- a/image-mirror.sh
+++ b/image-mirror.sh
@@ -91,6 +91,8 @@ function mirror_image {
   MANIFEST=$(skopeo inspect docker://${SOURCE}:${TAG} --raw)
   SCHEMAVERSION=$(jq -r '.schemaVersion' <<< ${MANIFEST})
   MEDIATYPE=$(jq -r '.mediaType' <<< ${MANIFEST})
+  SOURCES=()
+  DIGESTS=()
  
   # Most everything should use a v2 schema, but some old images (on quay.io mostly) are still on v1
   if [ "${SCHEMAVERSION}" == "2" ]; then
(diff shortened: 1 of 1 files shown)


Oats87 (dismissed):

image-mirror.sh
```diff
@@ -142,10 +142,10 @@ function mirror_image {
       ARCH=$(jq -r '.architecture' <<< ${CONFIG})
       DIGEST=$(jq -r '.config.digest' <<< ${MANIFEST})
       if grep -wqF ${ARCH} <<< ${ARCH_LIST}; then
-        if copy_if_changed "${SOURCE}:${TAG}" "${DEST}:${TAG}-${ARCH}" "${ARCH}"; then
-          echo -e "\tAdding ${DEST}:${TAG}-${ARCH} => ${DEST}:${TAG}"
-          docker buildx imagetools create --tag ${DEST}:${TAG} ${DEST}:${TAG}-${ARCH}
-        fi
+        copy_if_changed "${SOURCE}:${TAG}" "${DEST}:${TAG}-${ARCH}" "${ARCH}"
+        echo -e "\tAdding ${DEST}:${TAG}-${ARCH} => ${DEST}:${TAG}"
```
kinarashah, line 146: We now do this only if digests changed, so do we need these `echo Adding` lines? copy_if_changed logs all necessary info for both changed and unchanged.

brandond: Glad to get rid of it if nobody will miss it!

kinarashah (approved):

brandond:
@Oats87 you up for a final review on this?

Oats87 (approved):

PR #222, opened 2022-03-17
Bump coredns to 1.9.1

brandond:

#### Types of Change ####

version bump

#### Linked Issues ####

N/A


kinarashah (approved):

brandond:
@Crayeth that image looks fine to me; `failed to resolve rootfs` errors are usually a problem with your local image store, not the image on the registry itself.

Crayeth:
>that image looks fine to me; failed to resolve rootfs errors are usually a problem with your local image store, not the image on the registry itself.

@brandond you're probably right, I came here because of https://github.com/coredns/coredns/issues/5363 and thought maybe the wrong arch was mirrored into the rancher mirror.
However I found a workaround in https://github.com/abiosoft/colima/issues/219

PR #232, opened 2022-04-08
Add a daily image-mirror pipeline

cbron:
Solution for: https://github.com/rancherlabs/eio/issues/921

The goal of this PR is to add a pipeline step which will sync "latest" style images every day. They are in a separated list, and will run on push and [cron triggers](https://docs.drone.io/pipeline/docker/syntax/trigger/).

I ran the validator and daily steps locally, but I can't test for real without the credentials.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -10,7 +10,9 @@ steps:
   - name: validate-list
     image: library/ubuntu:21.04
     commands:
-      - grep -vE '^\s*(#|//)' images-list | sort -Vc
+    - grep -vE '^\s*(#|//)' images-list | sort -Vc
+    - grep -vE '^\s*(#|//)' images-list-daily | sort -Vc
+  # this step will run on every push
   - name: mirror-images
     image: rancher/dapper:v0.5.7
     volumes:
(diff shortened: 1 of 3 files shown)


images-list
```diff
@@ -447,8 +447,6 @@ minio/minio rancher/mirrored-minio-minio RELEASE.2020-07-13T18-09-56Z
 neuvector/controller rancher/mirrored-neuvector-controller 5.0.0-b1
 neuvector/enforcer rancher/mirrored-neuvector-enforcer 5.0.0-b1
 neuvector/manager rancher/mirrored-neuvector-manager 5.0.0-b1
-neuvector/scanner rancher/mirrored-neuvector-scanner latest
-neuvector/updater rancher/mirrored-neuvector-updater latest
```
cbron, line 451: This was just added this morning to do a single sync over, so it wasn't blocking the chart pr

brandond (approved):

.drone.yml
```diff
@@ -28,6 +30,25 @@ steps:
...
+      DOCKER_USERNAME:
+        from_secret: docker_username
+    commands:
+      - cat images-list-daily | dapper
+    settings:
+      custom_dns: 1.1.1.1
+    when:
+      event:
+      - push
+      - cron
```
superseb, line 51: When does this run exactly? I don't see a default or a configuration when it needs to run, but I'm also not familiar with Drone and cron triggers.

cbron: @paraglade setup a cron in drone that runs every day, that will trigger an event here.

superseb (approved):

PR #246, opened 2022-05-05
Bump skopeo/docker/buildx versions

superseb:
@cbron Who owns and maintains this repository? Do we have a test environment for it or should I test this change on my own account?

This bump should boost performance but I have other changes in queue (like seeing how we can reduce the 30 minutes build time further and/or move up processing the added images first)

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,13 +1,12 @@
-FROM quay.io/skopeo/stable:v1.1.1
+FROM quay.io/skopeo/stable:v1.7.0
 
 # Add jq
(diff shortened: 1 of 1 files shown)


cbron:
It's a group effort, no team owns it. No test env for it, though that would be a good idea. 
IMO for this repo I'd care more about security and stability than performance, any mistakes here can be dangerous, but we can go after both.

brandond (approved):

brandond:
Anyone up for rewriting the unholy bash script in go? Not sure if it'd actually speed up the image copy process much, but eliminating parsing and mangling the manifests with `jq` might realize some nice gains.

PR #272, opened 2022-08-19
Automate adding a tag to an existing image

superseb:
This is an attempt to automate adding tags to existing images. Currently this is time consuming (people have to wait for `CODEOWNERS`) and the expectations from the `CODEOWNERS` are all things that can be automated. This is only to automate tags to existing images in `images-list`, if the provided image(s) is/are not found in the list, it will exit.

The README changes should clarify how this works, if not, it needs to be improved.

This also includes some Dapper changes as I assumed we still wanted to use that model.

The GitHub Actions workflow can help people to make changes without having all the necessary tooling on their workstation or who are not familiar with the CLI. Currently the resulting PR still needs be approved (although it is no a dummy as this process should be fully checked, see the checks that it does in the README and in the scripts) so in my opinion the next step would be to (if a couple of the resulting PRs are satisfactory) to see if we can auto approve and auto merge these. The workflow also allows us to kick off these workflows via the API so we can even think of auto adding images when they get released (long term).

This PR is to gather feedback and of course if you see script issues, let me know.


superseb:
I see that Manuel is OOO, marking it ready for review to gather feedback.

manuelbuil (approved):

brandond:
I've never used inputs to drive actions; is there any way to test this short of merging it?

superseb:
It has a few parts where it interacts with GitHub which will fail without having a `GITHUB_TOKEN` set (pushing the branch and creating the PR), but you can probably test what you need using https://github.com/nektos/act;

```
git clone https://github.com/rancher/image-mirror 
cd image-mirror
git fetch origin pull/272/head:automate
git checkout automate             
curl https://raw.githubusercontent.com/nektos/act/master/install.sh | sudo bash
./bin/act workflow_dispatch -l
./bin/act -j add-tag-to-existing-image --env IMAGES=bla --env TAGS=v1.0.0
```

PR #291, opened 2022-09-28
Adds question about licenses to PRs

cbron:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,7 @@
 - [ ] New entries are in format `SOURCE DESTINATION TAG`
 - [ ] New entries are added to the correct section of the list (sorted lexicographically)
 - [ ] New entries have a repo created in Rancher Dockerhub (where the image will be mirrored to)
-- [ ] New entries have licenses of images listed here: TODO
+- [ ] New entries are licensed with Rancher favored licenses - Apache 2 and MIT - or approved licenses - as according to [CNCF approved licenses](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md).
 - [ ] Changes to scripting or CI config have been tested to the best of your ability
 
 #### Types of Change ####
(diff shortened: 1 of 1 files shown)


superseb (commented):
It would be nice to have guidance so the person creating the PR can already check what we do and do not allow. Also some newer people might have issues understanding what is being requested here. I'd rather have proper guidance and having the person filing it responsible for looking it up and making sure it complies than an unspecified question which slows the whole process down (unclear, back-and-forth etc)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,6 +6,7 @@
 - [ ] New entries are in format `SOURCE DESTINATION TAG`
 - [ ] New entries are added to the correct section of the list (sorted lexicographically)
 - [ ] New entries have a repo created in Rancher Dockerhub (where the image will be mirrored to)
+- [ ] New entries have licenses of images listed here: TODO
```
macedogm, line 9:
```suggestion
- [ ] New entries are licensed with Rancher [favored or approved licenses](https://github.com/rancherlabs/rancher-security/wiki/Rancher-Licensing-Standards)
```

macedogm (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
macedogm: Addressing @superseb comment.

brandond: Can we publish this somewhere externally-facing? This is a public repo. Hiding documentation supporting a checklist item in a private repo feels wrong.

cbron: I purposefully left this is a "list" question, as opposed to a checkbox question, so that people actually do the hard work of listing all licenses out. A checkbox is too easy to check and move on, we have seen lots of that already.
We could also talk about making our guidelines public, but that has wider implications.

macedogm: @brandond I don't think that the place to list the licenses was already defined. Nevertheless, I'm checking regarding listing the "favored or approved licenses", as you mentioned.

macedogm (approved):

PR #294, opened 2022-10-04
Add longhorn 1.3.2 images

PhanLe1010:
Signed-off-by: Phan Le <phan.le@suse.com>


PhanLe1010:
@innobead Do we need to add images like `longhornio/longhorn-manager rancher/longhornio-longhorn-manager ...` anymore? It seems that the our new charts doesn't use these images.

innobead:
> @innobead Do we need to add images like `longhornio/longhorn-manager rancher/longhornio-longhorn-manager ...` anymore? It seems that the our new charts don't use these images.

Yeah, we should remove them if they are not used anymore.

brandond (approved):

innobead (approved):

PR #328, opened 2022-12-09
Run max 1 pipeline at a time

superseb:
Only allow Drone to run the pipeline once at the same time, to avoid conflicting tasks. Also avoids the "wait to merge til the previous pipeline run completes" which I doubt even everyone knows.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,6 +2,9 @@
 kind: pipeline
 name: default-image-push-pull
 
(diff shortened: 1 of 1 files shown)


brandond (approved):

cbron (approved):

cbron:
Thanks. Should avoid any race condition problems now.

PR #332, opened 2022-12-22
Updated CPI image list

annablender:

- [ ] New entries are licensed with Rancher favored licenses - Apache 2 and MIT - or approved licenses - as according to [CNCF approved licenses](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md). - **what does this mean?**
- [ ] Changes to scripting or CI config have been tested to the best of your ability **NA**

#### Linked Issues ####

https://github.com/rancher/rancher/issues/39968

### DIFF
--- a/images-list
+++ b/images-list
@@ -115,12 +115,15 @@ gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provide
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.21.3
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.22.5
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.22.6
+gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.22.7
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.23.0
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.23.0-alpha.1
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.23.1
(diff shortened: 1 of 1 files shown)


images-list
```diff
@@ -115,12 +115,16 @@ gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provide
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.21.3
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.22.5
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.22.6
+gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.22.7
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.23.0
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.23.0-alpha.1
 gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.23.1
+gcr.io/cloud-provider-vsphere/cpi/release/manager rancher/mirrored-cloud-provider-vsphere-cpi-release-manager v1.23.2
```
kinarashah, line 122: Per issue description, we need the latest v1.23.3 - do we also need v1.23.2?

snasovich: I'm not aware why we would have to have v1.23.2 if we don't plan on referencing it. @annablender , do you know why we might need to have it?

annablender: Correct, I figured yes since it may be used for the constraints.. but perhaps not? I think I'm bumping straight to 1.23.3 here https://github.com/rancher/vsphere-charts/blob/c433c58733e91515157454e46310f8cb58732bea/charts/rancher-vsphere-cpi/values.yaml#L44 so 1.23.2 is not needed. @jiaqiluo Any thoughts?

snasovich: @annablender , constraints are for k8s versions. Since minor versions of CPI charts are aligned with minor versions of k8s they are for, it may be a bit confusing. But constraints are keyed off of k8s versions. :)

jiaqiluo: Yep, we should mirror 1.23.3 and skip 1.23.2

jiaqiluo (approved):

snasovich (approved):

Oats87 (approved):

PR #334, opened 2023-01-13
Automate adding a tag to an existing image

superseb:
This is for https://github.com/rancher/vsphere-charts/issues/41 and @manuelbuil also has asked for an automated way to provide images to be added.

This is an attempt to automate adding tags to existing images. Currently this is time consuming (people have to wait for CODEOWNERS) and the expectations from the CODEOWNERS are all things that can be automated. This is only to automate tags to existing images in images-list, if the provided image(s) is/are not found in the list, it will exit.

The README changes should clarify how this works, if not, it needs to be improved.

This also includes some Dapper changes as I assumed we still wanted to use that model.

The GitHub Actions workflow can help people to make changes without having all the necessary tooling on their workstation or who are not familiar with the CLI. Currently the resulting PR still needs be approved (although it is no a dummy as this process should be fully checked, see the checks that it does in the README and in the scripts) so in my opinion the next step would be to (if a couple of the resulting PRs are satisfactory) to see if we can auto approve and auto merge these. The workflow also allows us to kick off these workflows via the API so we can even think of auto adding images when they get released (longer term).


pjbgf (commented):
Maybe we could mention some of the new scripts (e.g. `scripts/add-tag-to-existing-image.sh`) on `.github/PULL_REQUEST_TEMPLATE.md` as using it would remove the need of some of those check boxes.

pjbgf (approved):
Quite nice the new workflow to allow triggering the process via GH. :+1: 

/lgtm

brandond (approved):

scripts/add-tag-to-existing-image.sh
```diff
@@ -0,0 +1,64 @@
...
+    exit 1
+  fi
+  if [ $MAPPING_COUNT -gt 1 ]; then
+    echo "More than one mapping found for image ${IMAGE}, checking for entry that contains '/mirrored-'"
+    MIRRORED_MAPPING=$(echo "${MAPPING}" | grep \/mirrored\-)
+    MIRRORED_MAPPING_COUNT=$(echo "${MIRRORED_MAPPING}" | wc -l)
+    if [ $MIRRORED_MAPPING_COUNT -ne 1 ]; then
+      echo "Unable to find mirrored mapping to use for image ${IMAGE}"
+      echo "MIRRORED_MAPPING: ${MIRRORED_MAPPING}"
+      echo "Unable to aueomatically add tag(s) for image ${IMAGE}, more than one mapping found (printed below), exiting"
```
brandond, line 33:
```suggestion
      echo "Unable to automatically add tag(s) for image ${IMAGE}, more than one mapping found (printed below), exiting"
```

brandond (commented):
nit

superseb:
@brandond I removed the option to feed images through stdin, because of the Makefile/dapper stuff, do you still want to keep this? The current PR only supports setting a file.

brandond:
As long as CI works I'm good with it, I mostly just used it because I wanted to be lazy and paste in an image list when testing.

brandond (approved):

superseb:
Ran some final test with @Anarkis and found some typos/misconfigurations in Drone configuration. We tested it using `drone exec` locally with a test org in Docker Hub.

.github/workflows/add-tag-to-existing-image.yml
```diff
@@ -0,0 +1,81 @@
...
+            echo "changes_exist=true" >> $GITHUB_ENV
+          fi
+      - name: Create branch, commit and push
+        if: ${{ env.changes_exist == 'true' }}
+        id: branch
+        run: |
+          BRANCH="githubaction-add-tag-$(date +%Y-%m-%d-%H-%M-%S)"
+          echo "branch=${BRANCH}" >> $GITHUB_OUTPUT
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
```
brandond, line 46: Should this come from the environment instead of being hardcoded?

superseb: The environment contains the actor info that is calling the workflow, these are the values we need to show the commits as coming from the actions actor.

brandond: Right, but if the environment contains the actor info, couldn't we just pass those env values into the git config, instead of hardcoding them here? I understand that they shouldn't change, but it seems more flexible/reusable to avoid hardcoding things that are controlled by GH and not us.

superseb: Just because it will never change, I don't see the value in making it configurable. It also introduces a dependency on EIO as now it is controlled outside the team's control (organization variable or per repo which then defeats the purpose of making it configurable)

I would also rather start saving time now by using this then having the perfect implementation, making this configurable is hardly any work and would be included in making it configurable for all repositories as we are already using this configuration in a few repositories (based on how we decide to do it)

brandond (approved):

PR #336, opened 2023-01-18
Add missing cilium-operator image

pjbgf:

#### Types of Change ####

This image is currently referenced by two charts but does not exist: 
https://github.com/rancher/rke2-charts/blob/83d8960812f14bb459d28eb873312ed3294dc180/packages/rke2-cilium/generated-changes/patch/values.yaml.patch#L119

https://github.com/rancher/rke2-charts/blob/main-source/packages/rke2-cilium-legacy/charts/values.yaml#L12-L13

#### Additional Notes ####

Added all existing cilium tags, to ensure users stuck on older version of the charts would be able to enable the operator, if they so wish.


superseb (commented):
@rbrtbnfgl Any info you have on this or was this just missed?

@pjbgf Are `v1.12.4` and `v1.12.5` the only ones actively in use? Do we not need earlier versions?

pjbgf:
It looks like it was just missed, the other variants `cilium/operator-{generic,aws,gke}` are all OK.

`v1.12.4` is currently in use (in the chart) and `v1.12.5` is what I am bumping it to. 
Alternatively we could pull all the versions the other cilium images as being used, just in case someone is caught on an older version of the chart.

pjbgf:
After some digging, I found other references to earlier versions:
https://github.com/rancher/rke2-charts/blob/main-source/packages/rke2-cilium-legacy/charts/values.yaml#L12-L13

Will update this PR accordingly.

rbrtbnfgl:
The RKE2 chart on the latest version is using the `1.12.4` version. We don't need to mirror the operator because the chart will use the generic image.

brandond:
@rbrtbnfgl if I am understanding your comment, we don't need to mirror these?

pjbgf:
I installed cilium on rke2 and it did pick up the generic version of the operator. 

@rbrtbnfgl I am assuming the same applies for the legacy chart: https://github.com/rancher/rke2-charts/blob/main-source/packages/rke2-cilium-legacy/charts/values.yaml#L12-L13

If so, feel free to just close the PR and I will update the other PR that refers to this.

brandond:
Yes, you can see where it builds the full image name at https://github.com/rancher/rke2-charts/blob/3fd16c7524207ccf52062d2dde318e706e8e1b2d/charts/rke2-cilium-legacy/rke2-cilium/1.11.101/charts/cilium/templates/cilium-operator/_helpers.tpl#L28-L31 - there is always some suffix on it.

pjbgf:
Nice one, thank you for checking I completely missed that. Closing this issue.

PR #365, opened 2023-03-24
Add scheduled workflow to retrieve tags for images

superseb:
This PR is to get feedback on this approach.

It adds a new scheduled workflow that runs a python script that can discover new tags for configured images based on a source, which will then trigger our current workflow that adds tags to existing images. This new workflow allows people to configure their images in `retrieve-image-tags/config.json` and the scheduled workflow will automatically dispatch the workflow that creates PRs for the inputs given to it. This allows teams to not to worry about;

- when a new image tag is released
- doing manual work when a new image tag is released

The README part should be self explanatory, if not, it needs to be adjusted so everyone understands how it works and how to use it.

retrieve-image-tags


thomasferrandiz (dismissed):

brandond:
It looks like this is using the GH release API to retrieve releases. Is there any other option for projects that don't use GH, or is that out of scope?

What about releases tags that aren't valid semver strings? Should we allow for filtering by a regex in addition to or instead of a version constraint?

What about projects that tag releases with a `v` prefix (v1.2.3) but tag images as `1.2.3` - do we need to allow for mutating or extracting the tag from the release?

What about projects that produce multiple variants of the release, and we want to select one, or select multiple? For example, we have:
```
// -alpine-21 variant
banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.11.5-alpine-21
// base and -debug variant
fluent/fluent-bit rancher/mirrored-fluent-fluent-bit 1.8.15
fluent/fluent-bit rancher/mirrored-fluent-fluent-bit 1.8.15-debug
```

superseb:
It will be an iterative approach to the logic, meaning that features that are needed will be added on demand. I am already working on different sources for versions (the various used registries).

We started with one example that are already helped with this approach (flannel) and people can start using it more and more, and if there is missing functionality, they or we can add it. The request is more to review the approach then to list all the possible use cases as there will be many more than this list, I'd rather help a few to start saving time now instead of having 100% coverage of everything from the start.

We have enabled issues for this repository recently and people can use that to request additional logic for their use case if needed.

brandond (approved):

Issue #399, opened 2023-05-24
Add check for target repository in CI


superseb:
We currently check if the source image and tag exists but not if the target repository exists. We should add that check so we don't have to rely on the person submitting the PR or on the reviewer to check it.

PR #400, opened 2023-05-24
Expand check script to check for target repository

superseb:
#### Types of Change ####

Add check if target repository exists, only if its a new target repository (this check is ignored if its not a new target repository)

#### Linked Issues ####

https://github.com/rancher/image-mirror/issues/399

### DIFF
--- a/scripts/check-new-images-exist.sh
+++ b/scripts/check-new-images-exist.sh
@@ -7,20 +7,36 @@ fi
 
 echo "Checking for new images in commit(s) against ${DIFF_CHECK}"
 
-NEW_IMAGES=$(git diff -U0 $DIFF_CHECK -- images-list | tail -n +5 | grep -v ^@@ | grep -v ^- | cut -d+ -f2 | awk '{ print $1":"$3 }')
+NEW_IMAGES=$(git diff -U0 $DIFF_CHECK -- images-list | tail -n +5 | grep -v ^@@ | grep -v ^- | cut -d+ -f2 | awk '{ print $1":"$3","$2 }')
 
 if [ -z "${NEW_IMAGES}" ]; then
   echo "Could not find new images in commit(s) against ${DIFF_CHECK}"
   exit 0
 fi
 
-echo "Found new images in commit(s) against ${DIFF_CHECK}: ${NEW_IMAGES}"
+echo -e "Found new images in commit(s) against ${DIFF_CHECK}:\n${NEW_IMAGES}"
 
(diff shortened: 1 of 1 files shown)


brandond (approved):

.drone.yml
```diff
@@ -20,6 +20,11 @@ steps:
     volumes:
       - name: docker
         path: /var/run/docker.sock
+    environment:
```
brandond, line 23: I will note that we don't usually give credentials to PR CI, only Merge. Is that OK?

superseb: Good point, I forgot about that. I removed the credentials part, we should be good with the rate limit as we don't run it that often.

scripts/check-new-images-exist.sh
```diff
@@ -7,20 +7,36 @@ fi
...
-  if ! skopeo inspect --retry-times=3 "docker://${NEW_IMAGE}" >/dev/null; then
-    echo "Image ${NEW_IMAGE} does not exist"
+  SOURCE_IMAGE=$(echo "${NEW_IMAGE}" | cut -d, -f1)
+  TARGET_IMAGE=$(echo "${NEW_IMAGE}" | cut -d, -f2)
+  if [ $(awk '$2 == "'"$TARGET_IMAGE"'" { print $2 }' images-list | wc -l) -eq 1 ]; then
+    echo "${TARGET_IMAGE} is a new target repository"
+    TARGET_NAMESPACE=$(echo "${TARGET_IMAGE}" | cut -d/ -f1)
+    TARGET_REPOSITORY=$(echo "${TARGET_IMAGE}" | cut -d/ -f2)
+    echo "Checking if Docker Hub namespace ${TARGET_NAMESPACE} and repository ${TARGET_REPOSITORY} exists"
+    if curl -H "Content-Type: application/json" --silent --fail "https://hub.docker.com/v2/namespaces/${TARGET_NAMESPACE}/repositories/${TARGET_REPOSITORY}/" > /dev/null; then
```
brandond, line 27: I think you want this to be an Accept header; the Content-Type header is part of the response.
```suggestion
    if curl -H "Accept: application/json" --silent --fail "https://hub.docker.com/v2/namespaces/${TARGET_NAMESPACE}/repositories/${TARGET_REPOSITORY}/" > /dev/null; then
```

superseb: Yeah I miss copy pasted it, but it doesn't really matter as we fail on response so I removed it.

brandond (approved):

PR #414, opened 2023-07-12
Add calico/cilium to auto PR configuration

superseb:
Adds configuration to automatically create PRs for calico and cilium required images

Because `quay.io/cilium/startup-script` uses a commit hash, I added `latest_entry` as a new config option to just return the first tag in the list.

@rbrtbnfgl can check if the configuration is good regarding source/destination and tags, output can be checked in https://github.com/rancher/image-mirror/actions/runs/5531007770/jobs/10091127096?pr=414#step:6:13


rbrtbnfgl:
You can skip the cilium startup-script. We are not using it and cilium moved all the setup inside the code the script is used only on older os.

rbrtbnfgl (approved):

PR #467, opened 2023-09-07
Add stale issue/PR workflow

superseb:
Add stale workflow similar to other `rancher` repositories


superseb:
@rancher/image-mirror-codeowners ^^

brandond (approved):

PR #468, opened 2023-09-07
bump kube-rbac-proxy

joshmeranda:

#### Types of Change ####

Version bump for kube-rbac-proxy `0.5.0` -> `0.14.0`

#### Linked Issues ####

https://github.com/rancher/rancher/issues/42019
https://github.com/rancher/charts/pull/2785

### DIFF
--- a/images-list
+++ b/images-list
@@ -207,6 +207,7 @@ gcr.io/google_containers/pause rancher/mirrored-pause 3.1
 gcr.io/google_containers/pause rancher/mirrored-pause 3.2
 gcr.io/kubebuilder/kube-rbac-proxy rancher/kube-rbac-proxy v0.5.0
 gcr.io/kubebuilder/kube-rbac-proxy rancher/mirrored-kube-rbac-proxy v0.5.0
+gcr.io/kubebuilder/kube-rbac-proxy rancher/mirrored-kube-rbac-proxy v0.14.0
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.12.4-alpine-1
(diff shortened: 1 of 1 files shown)


images-list
```diff
@@ -206,7 +206,9 @@ gcr.io/google_containers/k8s-dns-sidecar rancher/mirrored-k8s-dns-sidecar 1.15.1
 gcr.io/google_containers/pause rancher/mirrored-pause 3.1
 gcr.io/google_containers/pause rancher/mirrored-pause 3.2
 gcr.io/kubebuilder/kube-rbac-proxy rancher/kube-rbac-proxy v0.5.0
+gcr.io/kubebuilder/kube-rbac-proxy rancher/kube-rbac-proxy v0.14.0
```
brandond, line 209: do we need to keep mirroring new tags to the old un-mirrored repo? Can we update to the mirrored repo when switching to the new tag?

joshmeranda: Probably not, but I just assumed if both were included they were both in use and would need the update. Unless we have some charts referring to `rancher/kube-rbac-proxy:latest` that would need to have the newest versions, we can stop.

From the blame @rawmind0 added the original unmirrored image and @rmweir added the mirrored image. Maybe they have more context for why we have both in the image list / if we need to keep them updated.

joshmeranda: @brandond are you ok with only bumping the mirrored images moving forward? looks like thats what we do for the `rancher/bats-bats` and `rancher/mirrored-bats-bats`

brandond: yeah I think we should stop bumping tags for images without the mirrored- prefix.

joshmeranda: sick, fixed

brandond (approved):

PR #472, opened 2023-09-22
add images for monitoring rebase to prometheus 45.31.1

joshmeranda:


brandond (dismissed):

images-list
```diff
@@ -1254,6 +1259,7 @@ registry.k8s.io/git-sync/git-sync rancher/mirrored-git-sync v3.5.0
 registry.k8s.io/ingress-nginx/kube-webhook-certgen rancher/mirrored-ingress-nginx-kube-webhook-certgen v1.0
 registry.k8s.io/ingress-nginx/kube-webhook-certgen rancher/mirrored-ingress-nginx-kube-webhook-certgen v1.1.1
 registry.k8s.io/ingress-nginx/kube-webhook-certgen rancher/mirrored-ingress-nginx-kube-webhook-certgen v1.3.0
+registry.k8s.io/ingress-nginx/kube-webhook-certgen rancher/mirrored-ingress-nginx-kube-webhook-certgen v20221220-controller-v1.5.1-58-g787ea74b6
```
brandond, line 1262: this appears to be older than the current image by quite a bit; any idea why?

brandond (approved):

images-list
joshmeranda: I honestly don't. Looking through the history of the chart this is for doesn't make any clear indication of why the older image is used.

I think that the version the chart was using prior was needed as a workaround to an issue in the application's code, but once that issue was fixed they could go to a better maintained / more stable image

PR #476, opened 2023-10-03
updating kiali version from 1.72 -> 1.74

nicholasSUSE:

- [X] Confirm that you can pull the new images and tags from DockerHub

### DIFF
--- a/images-list
+++ b/images-list
@@ -1162,6 +1162,7 @@ quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.63.2
 quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.66.0
 quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.67.0
 quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.72.0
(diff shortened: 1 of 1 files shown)


images-list
```diff
@@ -1161,7 +1161,7 @@ quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.59.0
 quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.63.2
 quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.66.0
 quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.67.0
-quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.72.0
+quay.io/kiali/kiali rancher/mirrored-kiali-kiali v1.74.0
```
brandond, line 1164: See the PR checklist - you CANNOT remove existing tags. Add the new tag, without removing another one.

brandond (changes requested):

images-list
nicholasSUSE: Done

brandond (approved):

PR #482, opened 2023-10-11
Bump vsphere csi and cpi images and add config for auto-sync

thomasferrandiz:

- [ ] Change does not remove any existing Images or Tags in the images-list file
- [ ] Change does not remove / overwrite exiting Images or Tags in Rancher DockerHub
- [ ] New entries are licensed with Rancher favored licenses - Apache 2 and MIT - or approved licenses - as according to [CNCF approved licenses](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md).

#### Types of Change ####

- Version 
- Config update


brandond (approved):

retrieve-image-tags/config.json
```diff
@@ -73,6 +73,44 @@
...
+    "images": [
+      "registry.k8s.io/sig-storage/csi-provisioner"
+    ],
+    "versionSource": "github-latest-release:kubernetes-csi/external-provisioner"
+  },
+  "cpi-release-manager": {
+    "images": [
+      "gcr.io/cloud-provider-vsphere/cpi/release/manager"
+    ],
+    "versionSource": "github-latest-release:kubernetes/cloud-provider-vsphere/",
```
brandond, line 112: I think the trailing slash here is what's causing the script to crash
```suggestion
    "versionSource": "github-latest-release:kubernetes/cloud-provider-vsphere",
```

brandond (changes requested):

retrieve-image-tags/config.json
thomasferrandiz: Thanks!

retrieve-image-tags/config.json
```diff
@@ -73,6 +73,44 @@
...
+      "registry.k8s.io/sig-storage/csi-provisioner"
+    ],
+    "versionSource": "github-latest-release:kubernetes-csi/external-provisioner"
+  },
+  "cpi-release-manager": {
+    "images": [
+      "gcr.io/cloud-provider-vsphere/cpi/release/manager"
+    ],
+    "versionSource": "github-latest-release:kubernetes/cloud-provider-vsphere",
+    "versionFilter": "^v1\\.[0-9]+\\.[0-9]+$"
```
superseb, line 113: If you use `github-latest-release`, there won't be any filter applied as it will always just take the latest GitHub release. We can add the functionality if needed (for example, when multiple major/minor/patch are marked as latest whenever they release), but it isn't in there now.

thomasferrandiz: Ok then I think the best is to remove `github-latest-release` since we want to synchronize multiple images for different k8s versions and I don't know if they actually tag each one of them as latest at some point.

brandond: I suspect they just release a batch of new tags for multiple minors at the same time, similar to how we do our releases. Are you going to remove this in favor of matching tags from Docker Hub, or?

brandond (approved):

retrieve-image-tags/config.json
thomasferrandiz: I replaced it with 
```
    "versionSource": "github-releases:kubernetes/cloud-provider-vsphere",
    "versionFilter": "^v1\\.[2-9]+\\.[3-9]+$"
```
If I understand correctly, it should copy all the new images for the minors when they're published.

Issue #437, opened 2023-08-10
Automate adding images from chart as source


superseb:
The current implementation of automation is based on the "add tag to existing image" approach, meaning that you can provide images and tags and the logic will add if they do not already exists. As we only allow automated PRs for existing images, the logic will prevent from adding new images to the file (only add tags for images that are already present).

Later on logic was added to have a source of images/tags that can be looked up (for example, github releases or registries) and with that collected data, the workflow "add tag to existing images" is launched. 

The downside of the current approach is that we only allow for a list of images that require the same tag (usually from the same source). This means that if we have multiple images in a "release", we can't process that in a single request/PR. Currently we have multiple sources and images that all create separate PRs that make up one release.

For example, cilium uses quite a few different images with different release versions. The idea for this issue is to:

- Implement Helm chart as a new source for images/tags for a component (example project cilium, example source Cilium Helm chart)
- Retrieve Helm chart and use `helm template` with configurable values to get the correct rendered YAMLs
- Extract images from rendered YAMLs
- Feed the list of images + tags into a workflow to create the PR to add missing tags (we are still not adding new images)

Example mock up code:

```
# Lookup latest Helm chart from `index.yaml`, for this example hardcoded version
$ wget https://helm.cilium.io/cilium-1.14.0.tgz
# Helm template and extract images
$ helm template --set hubble.ui.enabled=true --set hubble.relay.enabled=true --set hubble.enabled=true --set azure.enabled=true --set envoy.enabled=true cilium-1.14.0.tgz  |  yq '..|.image? | select(.)' | sort -u | awk -F'@' '{ print $1 }'| sort -u
---
quay.io/cilium/cilium-envoy:v1.25.9-f039e2bd380b7eef2f2feea5750676bb36133699
quay.io/cilium/cilium:v1.14.0
quay.io/cilium/hubble-relay:v1.14.0
quay.io/cilium/hubble-ui-backend:v0.12.0
quay.io/cilium/hubble-ui:v0.12.0
quay.io/cilium/operator-azure:v1.14.0
# Launch workflow with comma separated list of images/tags to be processed (this is different from the current logic)
[...]

PR #489, opened 2023-10-16
Add Helm chart as source for retrieving images in workflow

superseb:
https://github.com/rancher/image-mirror/issues/437

Usage should be self explanatory from README.

It's not the cleanest solution as the initial design was to have an input for images and an input for tags and we would only add those tags to the images (so the same tags for the list of images). So for now I settled for wrapping the existing logic into a wrapper that takes a list of image + tag. This will help a lot of teams to not spend any more time on this (after initially adding the images for a first time).

This moves `cilium` images from the images and tags configuration to the new Helm configuration.

I tested this with quite a few Helm charts to make sure it won't break quickly:

```
{
  "epinio": {
    "versionSource": "helm-latest:https://epinio.github.io/helm-charts",
    "helmCharts": {
      "epinio/epinio": {
        "helmArgs": {
          "generic": [
            "global.domain=myepiniodomain.org"
          ]
        }
      }
    }
  },
  "kubewarden": {                                                                                                                                                                                                                                                                                                             
    "versionSource": "helm-latest:https://charts.kubewarden.io",                                                                                                                                                                                                                                                                      
    "helmCharts": {
      "kubewarden/kubewarden-controller": {},                                                                                                                                                                                                                                                                                 
      "kubewarden/kubewarden-defaults": {}                                                                                                                                                                                                                                                                                    
    }                                                                                                                                                                                                                                                                                                                         
  },
  "neuvector": {                                                                                                                                                                                                                                                                                                             
    "versionSource": "helm-latest:https://neuvector.github.io/neuvector-helm",
    "helmCharts": {
      "neuvector/core": {}
    }                                                                                                                                                                                                                                                                                                                         
  },
  "elemental": {                                                                                                                                                                                                                                                                                                              
    "versionSource": "helm-oci",                                                                                                                                                                                                                                                                                              
    "imageDenylist": [                                                                                                                                                                                                                                                                                                        
      "registry.suse.com/rancher/elemental-teal-channel"                                                                                                                                                                                                                                                                      
    ],                                                                                                                                                                                                                                                                                                                        
    "helmCharts": {                                                                                                                                                                                                                                                                                                           
      "oci://registry.suse.com/rancher/elemental-operator-chart": {}                                                                                                                                                                                                                                                          
    }                                                                                                                                                                                                                                                                                                                         
  }
}
```


superseb:
@rbrtbnfgl Is it okay to block the `quay.io/cilium/operator` image? I assumed as we are not using it currently, we don't want to add it now but it is being retrieved from the Helm chart.

The result of adding it as a Helm chart can be seen [here](https://github.com/rancher/image-mirror/actions/runs/6530334353/job/17729435402?pr=489#step:6:13) and would currently add the image `quay.io/cilium/startup-script:62093c5c233ea914bfa26a10ba41f8780d9b737f`

rbrtbnfgl:
we are using the cilium-operator image when we deploy cilium.

rbrtbnfgl:
We also can put the startup-script on the deny list. We removed it because it was only used on older K8s versions.

retrieve-image-tags/retrieve-image-tags.py
```diff
@@ -109,63 +192,92 @@ def _get_image_tags(registry, namespace, package, nexturl, page=1):
...
+                    shell=True,
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.STDOUT)
+            if 'imageDenylist' in values:
+                image_denylist = values['imageDenylist']
+            for chart in values['helmCharts']:
+                if 'helmArgs' in values['helmCharts'][chart]:
+                    for arg_key in values['helmCharts'][chart]['helmArgs']:
+                        helm_config_args = values['helmCharts'][chart]['helmArgs'][arg_key]
+                        helm_args = ["--set " + value for value in helm_config_args]
```
brandond, line 237: The name would suggest that you can pass through any helm arg, but this appears to be limited to just setting chart values. Should it be called `helmChartValues` instead?

retrieve-image-tags/retrieve-image-tags.py
```diff
@@ -73,26 +85,97 @@ def _get_image_tags(registry, namespace, package, nexturl, page=1):
...
+
+# Function to search textfile content
+def _image_tag_already_exist(filetext, image, tag):
+    matches = re.findall(fr"{image}\s.*?\s{tag}\n", filetext)
+    if matches:
+        return True
+    return False
+
+# Function to retrieve images from helm chart using helm template
+def _extract_unique_images_from_helm_template(repo_name, chart, helm_args=''):
```
brandond, line 117: We have several charts that use the kubernetes version to determine which images to use; making use of this requires setting `--kube-version` on the `helm template` command line. Could we add a field for that, similar to the `helmArgs` field that is used to set chart values?

brandond (changes requested):
nits/suggestions. this is a great feature though, I think it'll see a lot of use?

retrieve-image-tags/retrieve-image-tags.py
superseb: Agree, I changed the structure and naming for this and the `--kube-version` from below.

retrieve-image-tags/retrieve-image-tags.py
superseb: Done

superseb:
@rbrtbnfgl Added to denylist

rbrtbnfgl (approved):

README.md
```diff
@@ -106,25 +113,111 @@ See an example configuration below:
...
+          }
+        }
+      }
+    }
+  },
+  "kubewarden": {                                                                                                                                                                                                                                                                                                             
+    "versionSource": "helm-latest:https://charts.kubewarden.io",
+    "helmCharts": {
+      "kubewarden/kubewarden-controller": {},
+      "kubewarden/kubewarden-defaults": {}
```
brandond, line 174: a nit: This requires specifying charts as `$REPO/$CHART`, when `$REPO` is already the key for this entry in the dict. Charts should be specified as just the chart name; the tool can build the `$REPO/$CHART` arg required by the CLI from the dict key + the name of the chart.

```suggestion
      "kubewarden-controller": {},
      "kubewarden-defaults": {}
```

brandond (changes requested):

brandond (approved):

superseb:
@garyduan @selvamt94 @fgiudici @andreas-kupries We now have the option to add the Helm chart as source so when new Helm charts are published, new images can be retrieved from the chart and a PR will be created with the new images. See README for examples since this PR has been merged.

Issue #500, opened 2023-10-23
retrieve-image-tags : Support `--devel`, directory


andreas-kupries:
Ideas:

- Support `--devel` flag when accessing helm repositories to see dev charts as well.
- Support `directory` as a data source. The directory has to contain an unpacked helm chart.

The second idea would allows trialing the tool against charts in development and not released anywhere in no form.

superseb:
Linked PRs solve both requests.

PR #501, opened 2023-10-24
Add devel configuration option

superseb:
First part of https://github.com/rancher/image-mirror/issues/500

I went back and forth on this implementation, I started with a `devel` configuration option at first (like in this PR) but felt like maybe it could be more generic. So I transformed it into `helmArgs` but after implementing it, it didn't really make sense as it would not consume global Helm args but specific args for subcommands. And wether we need specific configuration directives for `helm template` and `helm show values`, we'll see in the future. 

For now,  this adds the option to use development versions (mainly used for local testing for now)

### DIFF
--- a/.github/workflows/validate-retrieve-image-tags-config.yml
+++ b/.github/workflows/validate-retrieve-image-tags-config.yml
@@ -2,7 +2,7 @@ name: Validate retrieve-image-tags config
 on:
   pull_request:
     paths:
-      - 'retrieve-image-tags/config.json'
+      - 'retrieve-image-tags/**'
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


brandond (approved):

.github/workflows/validate-retrieve-image-tags-config.yml
```diff
@@ -2,7 +2,7 @@ name: Validate retrieve-image-tags config
 on:
   pull_request:
     paths:
-      - 'retrieve-image-tags/config.json'
+      - 'retrieve-image-tags/**'
```
superseb, line 5: Added this as it wasn't running the validate workflow when anything else than `config.json` was changed

brandond (approved):

PR #527, opened 2023-11-24
added version constraint for etcd in retrieve image tags config

vardhaman22:

- [X] Change does not remove any existing Images or Tags in the images-list file
- [X] Change does not remove / overwrite exiting Images or Tags in Rancher DockerHub
- [ ] New entries are licensed with Rancher favored licenses - Apache 2 and MIT - or approved licenses - as according to [CNCF approved licenses](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md).

#### Types of Change ####

added version constraint for etcd image in retrieve images tags config

### DIFF
--- a/retrieve-image-tags/config.json
+++ b/retrieve-image-tags/config.json
@@ -131,6 +131,7 @@
     "images": [
       "quay.io/coreos/etcd"
     ],
-    "versionSource": "github-latest-release:etcd-io/etcd"
+    "versionSource": "github-releases:etcd-io/etcd",
+    "versionConstraint": ">3.5.10"
   }
 }


superseb (changes requested):
`github-latest-release` does not have `versionConstraint`, we only added it with the context of always using the latest release. We can add the functionality to `github-latest-release` or you can use `github-releases` instead.

brandond (approved):

vardhaman22:
> `github-latest-release` does not have `versionConstraint`, we only added it with the context of always using the latest release. We can add the functionality to `github-latest-release` or you can use `github-releases` instead.

thanks @superseb for pointing it out. 
updated

superseb (approved):

PR #544, opened 2023-12-26
[Improvement] Add a new data source, github-tagged-images-file, to automate retrieving images from a image-list file of a GitHub release

PhanLe1010:
### Current issue:

Longhorn would like to automate adding/mirroring images. However, the currently available list of data sources don't fit the use-case of Longhorn:
1. `github-releases` data source: This one finds new GitHub release tag and adds the images defined in the `images` field in the `config.json` file. For example:
    ```json
    {
      "vsphere-cpi": {
        "images": [
          "gcr.io/cloud-provider-vsphere/cpi/release/manager"
        ],
        "versionSource": "github-releases:kubernetes/cloud-provider-vsphere",
        "versionConstraint": ">1.21.0"
      }
    }
    ```
    This `config.json` file instructs the script to find GitHub release tags in the repo `kubernetes/cloud-provider-vsphere`. Then only add the `gcr.io/cloud-provider-vsphere/cpi/release/manager` image with the found tags to the [images-list](https://github.com/rancher/image-mirror/blob/master/images-list). This doesn't fit the use-case of Longhorn because the list of Longhorn images are not fixed. We added/removed images between the releases. Therefore, it would require manual works to modify the ` "images"` fied of the `config.json` oftenly 
1. `github-latest-release` data source: this data source has same limitation as the `github-releases`. Additionaly, Longhorn maintain multiple minor releases so a smaller version (e.g., `v1.4.5`) might be released after the current latest version (e.g., `v1.5.3`). This data source will not sync and add the smaller version (e.g., `v1.4.5`)
1. `registry` data source: Longhorn doesn't maintain a registry. Not applicable 
1. `helm-latest`, `helm-oci`, and `helm-directory` data sources. With these data sources, the workflow attempts to run `helm template` and extract the images from the workload (deployment/daemonset/pod) output of `helm template`. This approach doesn't work for Longhorn because not all Longhorn images appears in the output of `Helm template` (the images of Longhorn system managed components)

### Proposal

Add a new data source to automate retrieving images from file which contains the list of images of a GitHub release, `github-releases-images-file`. This will look up GitHub releases, excluding pre-releases, and find the list of images inside a specified file of the release. This can be used if your project maintains a list of images in a file, e.g., https://github.com/longhorn/longhorn/blob/master/deploy/longhorn-images.txt

An example of configuration for `github-releases-images-file` could be:
```json
{
  "longhorn": {
    "versionSource": "github-releases-images-file:longhorn/longhorn",
    "imagesFilePath": "deploy/longhorn-images.txt",
    "versionConstraint": ">=1.4.0"
  }
}
```

With the new `github-releases-images-file` data source, the above `config.json` instructs the GitHub workflow to:
1. Look up GitHub releases at the repo `longhorn/longhorn`, excluding pre-releases
1. Only consider the releases which are >=1.4.0
1. For each release, download the image list at `deploy/longhorn-images.txt` and add the newly found images to the [images-list](https://github.com/rancher/image-mirror/blob/master/images-list).

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ There is also a scheduled workflow called [Retrieve image tags](https://github.c
 
 - `github-releases`: This will use GitHub releases as source, excluding pre-releases. This can be used if you need to keep all tags from the configured images in sync with GitHub releases
 - `github-latest-release`: This will use the release on GitHub marked as `Latest release`. This can be used if you only want one release to be added that is marked as latest.
+- `github-tagged-images-file`: This will look up GitHub git repository tags, and find the list of images inside a specified file. The tag must have an associated release, with the pre-release flag unset. This can be used if your project maintains a list of images in a file, e.g., https://github.com/longhorn/longhorn/blob/master/deploy/longhorn-images.txt
 - `registry`: This will use the registry of the first image and look up available tags.
 - `helm-latest:helm-repo-fqdn`: This will add the helm-repo-fqdn, and use the latest version of configured Helm chart(s) (`helmCharts`) configured to extract the images. It uses `helm template` and `helm show values` to extract images. You can specify one ore more iterations of `helm template` by specifying one ore more `values` configurations to make sure all required images are extracted. If you want to block certain images from being extracted, you can use `imageDenylist` in the configuration. See example below.
 - `helm-oci`: This is the same as `helm-latest`, except you don't need to provide a repository but it will use the charts directly from the provided `helmCharts` (which should be formatted as `oci://hostname/chart`).
@@ -49,6 +50,9 @@ The current filters for tags are:
 - `latest`: Sorts the found tags numerically and returns only the latest tag
 - `latest_entry`: Returns the last found (newest) tag only (can be used when tags are not semver/cannot be sorted numerically)
 
+`github-tagged-images-file` specific options:
+- `imagesFilePath`: the path to the list of images inside a GitHub git repository
+
 Helm specific options:
 
 - `imageDenylist`: An array of images that will not be added (in case the image matching finds images that shouldn't be added as the automation only accounts for adding tags to existing images, not adding new images as they need to be approved first)
@@ -59,7 +63,7 @@ Helm specific options:
 
 See example configuration for `github-releases`, `github-latest-release` and `registry`:
 
-```
+```json
 {
   "vsphere-cpi": {
     "images": [
@@ -117,9 +121,20 @@ See example configuration for `github-releases`, `github-latest-release` and `re
 }
 ```
 
+See example configuration for `github-tagged-images-file`:
+```json
+{
+  "longhorn": {
+    "versionSource": "github-tagged-images-file:longhorn/longhorn",
+    "imagesFilePath": "deploy/longhorn-images.txt",
+    "versionConstraint": ">=1.4.0"
+  }
+}
+```
+
 See example configuration for `helm-latest:helm-repo-fqdn`:
 
-```
+```json
 {
   "cilium": {
     "versionSource": "helm-latest:https://helm.cilium.io",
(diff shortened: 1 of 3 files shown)


PhanLe1010:
cc @superseb

brandond (commented):
I like this idea, should it perhaps be named `github-release-list-artifact`, to indicate that the list comes from a release artifact?

PhanLe1010:
Thanks @brandond for reviewing. Currently, Longhorn doesn't include images list in release's artifact. It store the images list in a file in the source code of each release. Therefore, I think the datasource name `github-releases-images-file` might be a better fit. In the future, if some users have the images list in file in release's artifact, we can easily modify the code to for this use-case and the name `github-releases-images-file` still fits this use-case. WDYT?

brandond:
Ah that's interesting. Most Rancher projects make the airgap image list a release artifacts. This is a requirement for centralized image security scanning. Could LH consider doing the same?

PhanLe1010:
> This was a requirement for centralized image security scanning. Could LH consider doing the same?

I am not very sure about how Longhorn does image security scanning so not sure if this is something Longhorn would like to do. cc @innobead could you provide some thoughts?

Regarding to the data source naming, do you agree that `github-releases-images-file` sounds ok for both use-case (download from source code and download from release's artifact)? @brandond

PhanLe1010:
ping @brandond @superseb Could we continue the review?

brandond (changes requested):
I still don't think that "release" is the correct term to use here. It is not pulling content from GitHub Release Artifacts. It is is getting files from at git repo (or a GitHub repo, specifically) at a specific tag. Could we update things it to reflect that?

PhanLe1010:
Discussed with @brandond and we agree to change the name to `github-tagged-images-file`.
PR updated!

README.md
```diff
@@ -37,6 +37,7 @@ There is also a scheduled workflow called [Retrieve image tags](https://github.c
 
 - `github-releases`: This will use GitHub releases as source, excluding pre-releases. This can be used if you need to keep all tags from the configured images in sync with GitHub releases
 - `github-latest-release`: This will use the release on GitHub marked as `Latest release`. This can be used if you only want one release to be added that is marked as latest.
+- `github-tagged-images-file`: This will look up GitHub releases, excluding pre-releases, and find the list of images inside a specified file in the source code at the release's tag. This can be used if your project maintains a list of images in a file, e.g., https://github.com/longhorn/longhorn/blob/master/deploy/longhorn-images.txt
```
brandond, line 40:
```suggestion
- `github-tagged-images-file`: This will look up GitHub git repository tags, and find the list of images inside a specified file. The tag must have an associated release, with the pre-release flag unset. This can be used if your project maintains a list of images in a file, e.g., https://github.com/longhorn/longhorn/blob/master/deploy/longhorn-images.txt
```

README.md
```diff
@@ -49,6 +50,9 @@ The current filters for tags are:
 - `latest`: Sorts the found tags numerically and returns only the latest tag
 - `latest_entry`: Returns the last found (newest) tag only (can be used when tags are not semver/cannot be sorted numerically)
 
+`github-tagged-images-file` specific options:
+- `imagesFilePath`: the path to the list of images inside a GitHub release
```
brandond, line 54:
```suggestion
- `imagesFilePath`: the path to the list of images inside a GitHub git repository.
```

brandond (changes requested):
nits on the docs

README.md
PhanLe1010: Thank you. Fixed

README.md
PhanLe1010: Thank you. Fixed

brandond (approved):

PhanLe1010:
Hi @brandond all checks have passed and you have approved the PR. Is it ok to merge now?

PR #632, opened 2024-05-24
Add tag for fluentd image

dharmit:

#### Linked Issues ####

Addresses part of https://github.com/rancherlabs/image-scanning/issues/1366

### DIFF
--- a/images-list
+++ b/images-list
@@ -225,6 +225,7 @@ ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.13.3-alpine-
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.14.4-alpine-2
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.14.5-alpine-1
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.14.6-alpine-5
+ghcr.io/kube-logging/fluentd rancher/mirrored-banzaicloud-fluentd v1.16-ruby3.3-full-build.140
 ghcr.io/banzaicloud/logging-operator rancher/mirrored-banzaicloud-logging-operator 3.14.0
 ghcr.io/banzaicloud/logging-operator rancher/mirrored-banzaicloud-logging-operator 3.15.0
 ghcr.io/banzaicloud/logging-operator rancher/mirrored-banzaicloud-logging-operator 3.17.1
--- a/images-list
+++ b/images-list
@@ -225,7 +225,6 @@ ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.13.3-alpine-
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.14.4-alpine-2
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.14.5-alpine-1
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.14.6-alpine-5
-ghcr.io/kube-logging/fluentd rancher/mirrored-kube-logging-fluentd v1.16-ruby3.3-full-build.140
 ghcr.io/banzaicloud/logging-operator rancher/mirrored-banzaicloud-logging-operator 3.14.0
 ghcr.io/banzaicloud/logging-operator rancher/mirrored-banzaicloud-logging-operator 3.15.0
 ghcr.io/banzaicloud/logging-operator rancher/mirrored-banzaicloud-logging-operator 3.17.1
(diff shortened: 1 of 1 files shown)


dharmit:
@alexandreLamarre can you PTAL this PR and provide your feedback?

images-list
```diff
@@ -222,6 +222,7 @@ ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.13.3-alpine-
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.14.4-alpine-2
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.14.5-alpine-1
 ghcr.io/banzaicloud/fluentd rancher/mirrored-banzaicloud-fluentd v1.14.6-alpine-5
+ghcr.io/kube-logging/fluentd rancher/mirrored-kube-logging-fluentd v1.16-ruby3.3-full-build.140
```
brandond, line 225: any commentary on why we're changing the tag format, and pulling from a different upstream?

brandond: Also, you'll need to create and link a ticket for EIO to create the `rancher/mirrored-kube-logging-fluentd` repository.

brandond (changes requested):

images-list
dharmit: > any commentary on why we're changing the tag format, and pulling from a different upstream?

The image is no longer maintained under ghcr.io/banzaicloud. The change in tag is a mere reflection of how ghcr.io/kube-logging tags the image. 

> Also, you'll need to create and link a ticket for EIO to create the `rancher/mirrored-kube-logging-fluentd` repository.

Ack. Thanks for the tip. I'm hoping for someone from the O&B team to validate if I've made the correct changes before I perform the next steps.

mallardduck: My main question on the new tag we're targeting is what one we should use to be "safest".
I [tracked down a default image tag](https://github.com/kube-logging/logging-operator/blob/ace8f631482fa031077e2ff6319a11a7d1174904/pkg/sdk/logging/api/v1beta1/logging_types.go#L175
) string to here:
> "v1.16-full"

However, for our use-case I think tagging to a specific build (as you've done) is apt per their Best Practice:  

> To ensure that subsequent builds don't break your production environment, you may want to pin your image to a specific build:
>     
>     ghcr.io/kube-logging/fluentd:VERSION-IMAGE-TYPE-build.BUILD_NUMBER

In theory the 3.3 ruby based images will have w/e benefits Ruby 3.3 brings over 3.2.
However if this isn't the default being used upstream, I wonder if that puts us at more risk by switching?

dharmit: > However, for our use-case I think tagging to a specific build (as you've done) is apt per their Best Practice:
> 
> > To ensure that subsequent builds don't break your production environment, you may want to pin your image to a specific build:
> > ```
> > ghcr.io/kube-logging/fluentd:VERSION-IMAGE-TYPE-build.BUILD_NUMBER
> > ```

This was the exact reason I went with a tag that looked kinda unreadable.



> if this isn't the default being used upstream, I wonder if that puts us at more risk by switching?

For the same reason that you mentioned above, IMO, we should use an image with build number.

mallardduck: Sorry to clarify - I think we're on the same page for targeting build specific tags.
I was less clear about what version "upstream" ships by default to users.
Are they using a Ruby 3.2 or Ruby 3.3 based image/fluentd build for `"v1.16-full"`?

My logic being that we probably want to use whatever version is most tested and used by end-users of upstream. That way we (in theory) have a lower risk of running into unexpected edge case issues. Or maybe I'm over thinking that part of things.

Maybe @joshmeranda has some ideas on this as he did a lot of hard work around the initial rebase to the new logging stuff.

ericpromislow:
Doesn't look like you need another reviewer, but I can review that the ghrc.io image has no HIGH or CRITICAL CVEs

ericpromislow (changes requested):
Need to move the line starting with `ghcr.io/kube-logging/fluentd rancher/mirrored-kube-logging-fluentd` to the line after `ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.10.0` to pass the sort check.

images-list
dharmit: > Or maybe I'm over thinking that part of things.

It's just our occupational hazard, I'd wager. 😉

dharmit:
> Need to move the line starting with `ghcr.io/kube-logging/fluentd rancher/mirrored-kube-logging-fluentd` to the line after `ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.10.0` to pass the sort check.

Thanks, @ericpromislow. I thought images were grouped based on repository. Didn't know they were alphabetically sorted. :)

brandond:
lgtm; still need the new target repo created by EIO

dharmit:
> lgtm; still need the new target repo created by EIO

@brandond done. Can you PTAL when you have a chance?

mallardduck (approved):
:shipit:

brandond (approved):

Issue #666, opened 2024-07-10
Drone Publish Jobs not Running


mallardduck:
The drone publish jobs for this repo are no longer running, as such we are missing the images that should have been mirrored when this https://github.com/rancher/image-mirror/pull/664 was merged. Without those images this is a blocker for https://github.com/rancher/rancher/pull/46082

It appears that `drone-publish` for this repo is inactive: https://drone-publish.rancher.io/rancher/image-mirror/settings

mallardduck:
Just have one more PR related to this that I'd like to follow up on (merge or close) before closing this out.
https://github.com/rancher/image-mirror/pull/670

mallardduck:
I've updated the PR so this can close once it is merged.

PR #668, opened 2024-07-11
Verify prs with GitHub Action

thardeck:
#### Types of Change ####

Add GitHub Action to verify prs similarly to Drone.

#### Additional Notes ####

Updated the README.md accordingly and removed a leftover file from the Dapper Docker image.

PR #670, opened 2024-07-11
Remove dapper from GHA calls

mallardduck:
#### Types of Change ####

Removes dapper from GHA workflows by calling the scripts directly.

#### Linked Issues ####

Follow up for:
https://github.com/rancher/image-mirror/pull/668
https://github.com/rancher/image-mirror/pull/669

Will fix: https://github.com/rancher/image-mirror/issues/666

### DIFF
--- a/.github/workflows/add-tag-to-existing-image.yml
+++ b/.github/workflows/add-tag-to-existing-image.yml
@@ -32,10 +32,10 @@ jobs:
         uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938 # v4.2.0
       - name: Run add-tag-to-existing-image.sh
         if: github.event.inputs.images != '' && github.event.inputs.tags != ''
-        run: make add-tag-to-existing-image.sh
+        run: ./scripts/add-tag-to-existing-image.sh
       - name: Run add-full-image-wrapper.sh
         if: github.event.inputs.full_images != ''
-        run: make add-full-image-wrapper.sh
+        run: ./scripts/add-full-image-wrapper.sh
       - name: Check for repository changes
         run: |
           if git diff --name-only --exit-code; then


brandond (approved):

brandond:
I still see a drone check on this PR, does it need to be rebased?

mallardduck:
@brandond - yeah that's odd, seems like the repo settings may have had Drone PR added back?

brandond:
It's at least not blocking the merges any more, so I can go ahead and merge this if you'd like.

mallardduck:
I'll let @thardeck give this a review first and then after that lets merge it. I like the collaboration we're doing and think having them look it over and giving a review too will be good.

mallardduck:
Ahh - yeah I was still on a stale `master` branch. Looks like I had never added an upstream to pull in the new changes and was on my out of sync fork still. Fixed that now and will adjust readme as mentioned.

I'm also looking into the add-tag job issue and will probably make a new PR for that. I will also look into addressing a way to clean up the `gha-` branches via workflow too, since I'm seeing a lot of those on the repo.

mallardduck:
So it seems our failing `add-tag-to-existing-image` like: https://github.com/rancher/image-mirror/actions/runs/9905573716/job/27365382388

Are being caused by an upstream k8s issue caused by an upstream (to them) issue with microsoft base images they use; based on this issue: https://github.com/kubernetes-csi/livenessprobe/issues/273

Also just noticed the [release note](https://github.com/kubernetes-csi/livenessprobe/releases/tag/v2.13.0) says they cannot build windows images; confused why this caused them to not publish linux images, but likely related.

mallardduck:
Wonder if we should merge this PR or close it? I know some folks still like the uniformity Dapper provides, so maybe we don't remove dapper? I've enjoyed keeping it around in a repo where I need to run the CI scripts locally. I'm not sure if that applies to this repo though, so maybe removing it is fine since we don't need that local use case?

brandond:
I am +1 on dropping it from the workflows, but I do think there is value in leaving artifacts in the repo for devs that still use dapper for local testing - at least until we receive direction from management that we are to do away with it entirely.

mallardduck:
@brandond - this should be good to review now, scoped down to just remove dapper from CI running. However it remains for local dev usage.

brandond (approved):

joshmeranda (approved):

colstrom (approved):

brandond:
yolo

PR #687, opened 2024-08-02
bump image for monitoring rebase

joshmeranda:

- [x] If updating an existing entry, verify the `SOURCE` is still accurate and upstream hasn't been migrated to a new registry or repo (if they've migrated, a new repo request to EIO is needed to comply with the `SOURCE DESTINATION TAG` pattern)

#### Types of Change ####

New image tags


images-list
```diff
@@ -220,6 +220,7 @@ gcr.io/google_containers/k8s-dns-sidecar rancher/mirrored-k8s-dns-sidecar 1.15.1
 gcr.io/google_containers/pause rancher/mirrored-pause 3.1
 gcr.io/google_containers/pause rancher/mirrored-pause 3.2
 gcr.io/kubebuilder/kube-rbac-proxy rancher/kube-rbac-proxy v0.5.0
+gcr.io/kubebuilder/kube-rbac-proxy rancher/mirrored-kubebuilder-kube-rbac-proxy v0.18.0
```
brandond, line 223: why are we changing the target from `rancher/mirrored-kube-rbac-proxy` to `rancher/mirrored-kubebuilder-kube-rbac-proxy`? I think the target should remain as is, and v0.18.0 should go at the bottom of the list just below this line.

If the target change is intentional, can you confirm that the new repo exists and the bot account can push to it?

brandond (changes requested):

images-list
joshmeranda: Your right, didn't catch this @brandond

images-list
```diff
@@ -223,6 +223,7 @@ gcr.io/kubebuilder/kube-rbac-proxy rancher/kube-rbac-proxy v0.5.0
 gcr.io/kubebuilder/kube-rbac-proxy rancher/mirrored-kube-rbac-proxy v0.5.0
 gcr.io/kubebuilder/kube-rbac-proxy rancher/mirrored-kube-rbac-proxy v0.14.0
 gcr.io/kubebuilder/kube-rbac-proxy rancher/mirrored-kube-rbac-proxy v0.15.0
+gcr.io/kubebuilder/kube-rbac-proxy rancher/mirrored-kube-rbac-proxy v0.18.0
```
brandond, line 226: I don't see anything in upstream newer than v0.16.0. Is the upstream project now publishing somewhere else?

joshmeranda: yeah I've been looking into that. Looks like the upstream chart changed their upstream to `brancz/kube-rbac-proxy` which wasn't clear when I was rebasing the chart. Will need a new repo

brandond: Looking at the repo, I suspect that the gcr.io/kubebuilder/kube-rbac-proxy image was always just a mirror of the `brancz/kube-rbac-proxy` image since the project still hasn't been accepted into the kubernetes org. Kind of an odd situation.

brandond (approved):

brandond:
merged. You might check on the new repo in a bit to make sure the tags get pushed successfully.

PR #706, opened 2024-08-26
include fluentd latest base build

jbiers:

#### Types of Change ####

New image tag

#### Linked Issues ####

[#1366](https://github.com/rancherlabs/image-scanning/issues/1366)

### DIFF
--- a/images-list
+++ b/images-list
@@ -264,8 +264,8 @@ ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.6.2
 ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.8.1
 ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.9.0
 ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.10.0
-ghcr.io/kube-logging/fluentd rancher/mirrored-kube-logging-fluentd v1.16-ruby3.3-full-build.140
 ghcr.io/kube-logging/fluentd rancher/mirrored-kube-logging-fluentd v1.16-base-build.142
+ghcr.io/kube-logging/fluentd rancher/mirrored-kube-logging-fluentd v1.16-ruby3.3-full-build.140
 ghcr.io/kube-logging/logging-operator rancher/mirrored-kube-logging-logging-operator 4.4.0
 ghcr.io/kube-logging/logging-operator rancher/mirrored-kube-logging-logging-operator 4.8.0
 ghcr.io/kube-vip/kube-vip-iptables rancher/mirrored-kube-vip-kube-vip-iptables v0.6.0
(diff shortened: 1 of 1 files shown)


mallardduck (changes requested):
This LGTM overall - however I have some minor feedback.

We probably want to match the versions of the images with that of upstream.
So we would take the version we mirror from here:

https://github.com/rancher/charts/blob/1cde28a5180fdea1db8fa35dbf1eae73f9bf9547/packages/rancher-logging/package.yaml#L1

And go find the equivalent tag (in this case 4.8.0) on the repo's Github, and find the default they set for the tag version: https://github.com/kube-logging/logging-operator/blob/f8081b7bdbd2622820d2384fbe86b58873dc8db2/pkg/sdk/logging/api/v1beta1/logging_types.go#L175

In this case we see it is set to `v1.16-4.8-full` which is their rolling release tag, so we have to use that to resolve the static tag since this will change. And the current one seems to resolve to [v1.16-4.8-full-build.149](https://github.com/orgs/kube-logging/packages/container/fluentd/245314979?tag=v1.16-4.8-full-build.149)

So that's the value we should use to mirror this image.

Note - might be worth considering what other images we mirror for logging and if we are using the expected versions. Taking steps to properly mirror those might be good too. (i.e. we use rancher/mirrored-jimmidyson-configmap-reload, but should/could probably mirror the same one upstream uses `ghcr.io/kube-logging/config-reloader`)

brandond:
This change LGTM - it sounds like there is some work necessary to pin the specific tag used by the chart though? I'll wait for that conversation to resolve before approving and merging.

jbiers:
> This LGTM overall - however I have some minor feedback.
> 
> We probably want to match the versions of the images with that of upstream. So we would take the version we mirror from here:
> 
> https://github.com/rancher/charts/blob/1cde28a5180fdea1db8fa35dbf1eae73f9bf9547/packages/rancher-logging/package.yaml#L1
> 
> And go find the equivalent tag (in this case 4.8.0) on the repo's Github, and find the default they set for the tag version: https://github.com/kube-logging/logging-operator/blob/f8081b7bdbd2622820d2384fbe86b58873dc8db2/pkg/sdk/logging/api/v1beta1/logging_types.go#L175
> 
> In this case we see it is set to `v1.16-4.8-full` which is their rolling release tag, so we have to use that to resolve the static tag since this will change. And the current one seems to resolve to [v1.16-4.8-full-build.149](https://github.com/orgs/kube-logging/packages/container/fluentd/245314979?tag=v1.16-4.8-full-build.149)
> 
> So that's the value we should use to mirror this image.
> 
> Note - might be worth considering what other images we mirror for logging and if we are using the expected versions. Taking steps to properly mirror those might be good too. (i.e. we use rancher/mirrored-jimmidyson-configmap-reload, but should/could probably mirror the same one upstream uses `ghcr.io/kube-logging/config-reloader`)

I appreciate the thorough explanation, thanks! Just updated the image tag to match the expected one. Also, I'll open a separate issue to further investigate the use of mirrored images for logging as suggested.

mallardduck (approved):
🚢

brandond (approved):

jbiers:
@brandond I might need some help with merging this when you're available :D

Also, the [DockerHub repo](https://hub.docker.com/r/rancher/mirrored-kube-logging-fluentd) for this image seems to not have any deployed tags even though we already had one in entry for it in the image-list file. Do you know what might be the reason?

brandond:
https://github.com/rancher/image-mirror/actions/runs/10574628248/job/29296558931#step:8:3621
```
time="2024-08-27T09:02:52Z" level=fatal msg="writing blob: initiating layer upload to /v2/rancher/mirrored-kube-logging-fluentd/blobs/uploads/ in registry-1.docker.io: errors:\ndenied: requested access to the resource is denied\nunauthorized: authentication required\n"
===
Failed copying image for rancher/mirrored-kube-logging-fluentd
===
```

It looks like EIO has not granted the bot account access to publish to that repo. That should probably be fixed before we add additional tags.

cc @dharmit 
* https://github.com/rancherlabs/eio/issues/2369

PR #718, opened 2024-09-05
Add tags for various O&B team bumps

mallardduck:

### DIFF
--- a/images-list
+++ b/images-list
@@ -49,6 +49,7 @@ curlimages/curl rancher/mirrored-curlimages-curl 7.73.0
 curlimages/curl rancher/mirrored-curlimages-curl 7.77.0
 curlimages/curl rancher/mirrored-curlimages-curl 7.83.1
 curlimages/curl rancher/mirrored-curlimages-curl 7.85.0
+curlimages/curl rancher/mirrored-curlimages-curl 8.9.1
 directxman12/k8s-prometheus-adapter rancher/mirrored-directxman12-k8s-prometheus-adapter v0.8.3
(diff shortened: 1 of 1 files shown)


jbiers (approved):

images-list
```diff
@@ -270,6 +271,7 @@ ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.6.2
 ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.8.1
 ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.9.0
 ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.10.0
+ghcr.io/jimmidyson/configmap-reload rancher/mirrored-jimmidyson-configmap-reload v0.13.1
```
brandond, line 274: just to confirm, we are no longer sourcing this from `docker.io/jimmidyson/configmap-reload`

brandond (approved):

images-list
mallardduck: Correct - same git repo source, but they stopped pushing to dockerhub and started using GHCR. I think in a scenario like that we can keep the same mirror destination right?

Note from upstream:
> Since version v0.10, the Docker image is available from ghcr.io at https://github.com/jimmidyson/configmap-reload/pkgs/container/configmap-reload. Previous versons are available from Docker Hub at https://hub.docker.com/r/jimmidyson/configmap-reload

brandond: yeah sgtm

PR #720, opened 2024-09-05
add new config-reloader image

jbiers:

#### Types of Change ####

New image to be used by rancher-logging.
The logging-operator has moved from https://github.com/jimmidyson/configmap-reload in favor of https://github.com/kube-logging/config-reloader, as can be seen [here](https://github.com/kube-logging/logging-operator/blob/f8081b7bdbd2622820d2384fbe86b58873dc8db2/pkg/sdk/logging/api/v1beta1/logging_types.go#L172). We need to keep our image dependencies up to date with upstream.

#### Linked Issues ####
https://github.com/rancher/rancher/issues/46924

### DIFF
--- a/images-list
+++ b/images-list
@@ -272,6 +272,7 @@ ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.8.1
 ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.9.0
 ghcr.io/epinio/epinio-unpacker rancher/mirrored-epinio-epinio-unpacker v1.10.0
 ghcr.io/jimmidyson/configmap-reload rancher/mirrored-jimmidyson-configmap-reload v0.13.1
+ghcr.io/kube-logging/config-reloader rancher/mirrored-kube-logging-config-reloader v0.0.5
 ghcr.io/kube-logging/fluentd rancher/mirrored-kube-logging-fluentd v1.16-4.8-full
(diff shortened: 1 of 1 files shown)


brandond:
so now we're using two different images? ref: https://github.com/rancher/image-mirror/pull/718/commits/c72b02c7e38f62dc792fd49a5edd3457435dd764
cc @mallardduck

jbiers:
> so now we're using two different images? ref: [c72b02c](https://github.com/rancher/image-mirror/commit/c72b02c7e38f62dc792fd49a5edd3457435dd764) cc @mallardduck

The dependency @mallardduck added was for the monitoring chart, while this one is for logging. Probably each upstream project has moved on different directions regarding this project, but correct me if I'm mistaken ;)

mallardduck:
> so now we're using two different images? ref: #718 / [c72b02c](https://github.com/rancher/image-mirror/commit/c72b02c7e38f62dc792fd49a5edd3457435dd764) cc @mallardduck

Yeah, so I totally get the confusion since they are very similar things in both name and function. However as @jbiers said, this one comes from upstream logging and the one from my PR comes from upstream monitoring (specifically for alerting drivers charts). It seems that logging used to use the same one, but then they switched to one they maintain.

It does make me wonder if we could converge onto using the same thing in both places, however given the trajectory of the monitoring project I'm not sure it's worth the dev/QA cycles to change that much.

brandond:
Looks like you need to open an EIO ticket to create the new target repo. Make sure to ask for the bot account to have write access.

jbiers:
@brandond the eio issue has now been resolved :)

brandond (approved):

brandond:
you might check and make sure the tag is successfully mirrored once CI finishes

PR #829, opened 2025-02-06
Introduce `regsync`, Go scripts, `config.yaml` and `regsync.yaml`

adamkpickering:
This PR takes the first steps towards overhauling `rancher/image-mirror` in order to accomplish the goals set out by [EDR 005](https://docs.google.com/document/d/1IVu4Lq13yKiSs0gl_tIJ3vuidVmt0FIkO2mlZXTRAH4/edit?pli=1&tab=t.0). Here is the [JIRA issue](https://jira.suse.com/browse/SURE-9686). This PR:

- Adds Go scripts with `generate-regsync` subcommand that generate `regsync.yaml` (the `regsync` config) from `config.yaml`. These scripts will facilitate the writing of unit tests, automation, and better validation.
- Updates `README.md` to reflect the current state of the repo.

The idea is to develop the new scripts and setup alongside the old setup. I want to gradually migrate configuration from `images-list` and `images-list-daily` to `config.yaml` and `regsync.yaml`. Once that is done, `retrieve-image-tags/config.json` and the associated scripts can also be migrated. The old setup should never stop working until we are ready to completely migrate to the new setup.


main.go
brandond, line 1: nit: should new tools perhaps live in a `hack` or `cmd` subdirectory, instead of at the root of the repo? We have `scripts` for more recent shell-based additions.

adamkpickering: Hi @brandond! I'm glad you commented - I think you've done a bunch of work on this repo in the past and no doubt know more than me. If you are able to give this PR a more thorough review I'm sure it would be productive... no pressure though of course :smile:

As to your comment, I'm curious as to why? Has this bitten you in the past?

I was thinking that `go run main.go` would be the gateway to all scripting functionality eventually, but maybe that isn't as tidy as it could be. We could move all of the Go code (`main.go`, `pkg/`, `go.mod`, `go.sum`) into a `cmd/` (or whatever) directory, and the only cost would be having to write `go run cmd/main.go ...` instead of `go run main.go ...`. I don't *think* this would cause any issues when running `go test`, `golangci-lint` etc.

On the other hand, there are already `Dockerfile.dapper`, `Makefile`, and `entrypoint.sh`, which are arguably source code...

mallardduck: > On the other hand, there are already Dockerfile.dapper, Makefile, and entrypoint.sh, which are arguably source code...

Sidenote, because Dapper is not something with any maintainers we could/should probably even consider removing that at some point. I don't have strong feelings on if that should be done with this effort, or afterwards - I could see either one making sense.

adamkpickering: I would like to remove `dapper`. However, I don't think it is a good idea to do that while we are using the current scripts. `dapper` is necessary for the Python scripts, assuming we don't want to mess around with `venv`s et cetera. `dapper` might be needed for the shell scripts as well.

I intend to replace the current scripts with Go and `regsync`. Part of the reason for this is to remove dapper - if we are using Go, we don't need to worry (much) about runtime dependencies.

I'll take a look at putting the Go scripts in their own directory :+1:

README.md
```diff
@@ -3,7 +3,72 @@
 This repo is dedicated to mirror images from other organizations into Rancher.
 There are no packaging changes or changes in the layers of these images.
 
-## Mirroring images
+> [!IMPORTANT]
+> This repository is currently undergoing an overhaul in order to accomplish the
+> goals laid out in
+> [EDR 005](https://docs.google.com/document/d/1IVu4Lq13yKiSs0gl_tIJ3vuidVmt0FIkO2mlZXTRAH4/edit?pli=1&tab=t.0#heading=h.k15gkmaregsk).
```
diogoasouza, line 9: Should we have a private document linked here?

main_test.go
```diff
@@ -0,0 +1,55 @@
+package main
```
diogoasouza, line 1: I may have missed it, but do we run tests in the CI currently?

pkg/regsync/config.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// Config is parsed configuration file for regsync
+type Config struct {
+	Version int          `json:"version,omitempty"`
+	Creds   []ConfigCred `json:"creds"`
+	Sync    []ConfigSync `json:"sync"`
+}
+
+type ConfigCred struct {
```
diogoasouza, line 27: I think you missed adding documentation to this struct

main.go
```diff
@@ -0,0 +1,98 @@
...
+			},
+		},
+	}
+
+	if err := app.Run(os.Args); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func generateRegsyncYaml(ctx *cli.Context) error {
```
diogoasouza, line 35: Can you maybe add a short description of that this does here?

main.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if err := regsync.WriteConfig(regsyncYamlPath, regsyncYaml); err != nil {
+		return fmt.Errorf("failed to write %s: %w", regsyncYamlPath, err)
+	}
+
+	return nil
+}
+
+func getRegsyncEntries(repo config.Repository, image config.Image) ([]regsync.ConfigSync, error) {
```
diogoasouza, line 73: Can you maybe add a short description of that this does here?

main_test.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/rancher/image-mirror/pkg/config"
+	"github.com/rancher/image-mirror/pkg/regsync"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetRegsyncEntries(t *testing.T) {
```
diogoasouza, line 12: More of a nit/personal preference:
I think it's clearer when we have the input and output of the test near each other, usually in a struct i.e. https://github.com/rancher/rancher/blob/23edf6f5a47f7676cf7ce065c51520ffe23f4b50/pkg/git/git_test.go

README.md
adamkpickering: @diogoasouza I asked @prachidamle to weigh in here.

@rohitsakala the EDR is indeed not marked approved. But this is not because we haven't done the work of vetting the proposal. I've gone through the process of holding a meeting, responding to comments, et cetera; all stakeholders have given their input and changes have been made accordingly. I then reached out to a variety of people in order to try to get the EDR marked approved before opening PRs, but that did not go anywhere. Because of that, and because these changes do not break anyone else's projects, I'm moving forward. If there are any problems then I will no doubt hear about them.

prachidamle: Yes we should not merge untill EDR is approved - this should treated more as a POC until then.

prachidamle: About private docs - lets not link any internal docs please, just like with any JIRA issues. @adamkpickering

main_test.go
adamkpickering: At this time there is no CI for the Go code. I have written the GHA workflows for that as part of another PR. Though, it may make more sense to add them here. Let me know if you want me to do that.

main.go
```diff
@@ -0,0 +1,98 @@
...
+	"log"
+	"os"
+	"strings"
+
+	"github.com/rancher/image-mirror/pkg/config"
+	"github.com/rancher/image-mirror/pkg/regsync"
+	"github.com/urfave/cli/v2"
+)
+
+const configYamlPath = "config.yaml"
```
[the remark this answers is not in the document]
adamkpickering: `config.yaml` and the Go code are located in the same repository, and we will always be running the Go code from the root of this repository, so there isn't much point in doing this. I can add it if you want though - maybe you have a use case I haven't thought of.

regsync.yaml
```diff
@@ -0,0 +1,44 @@
+##################################################
```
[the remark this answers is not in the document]
adamkpickering: I wasn't planning on removing `config.yaml` or `regsync.yaml` in this PR. I plan on replacing them with actual values in the first PR that migrates some values from `images-list` to `config.yaml` (and thus `regsync.yaml`).

adamkpickering:
> I guess more PRs will follow to extend/improve this.

Yes, that's the case. If you're curious what that will look like, you can access the epic from the linked jira issue, which details planned future work.

main.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+			syncEntries, err := getRegsyncEntries(repo, image)
+			if err != nil {
+				return fmt.Errorf("failed to convert Image with SourceImage %q: %w", image.SourceImage, err)
+			}
+			regsyncYaml.Sync = append(regsyncYaml.Sync, syncEntries...)
+		}
+	}
+
+	if err := regsync.WriteConfig(regsyncYamlPath, regsyncYaml); err != nil {
```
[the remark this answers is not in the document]
adamkpickering: I have been looking at validation as a separate piece of work (see [SURE-9690](https://jira.suse.com/browse/SURE-9690)). Validation is certainly an important part of the picture, but I wanted to keep the scope of this PR from getting too big.

For what it's worth, the scripts and the config are in the same repo, so if we need to tweak the scripts as we're making a config change, we can do that. This is a different way of thinking about the problem compared to what we're used to i.e. rancher/charts and rancher/partner-charts.

If you disagree with treating validation as a separate work item, I am ok with adding it.

main.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+	}
+
+	if err := app.Run(os.Args); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func generateRegsyncYaml(ctx *cli.Context) error {
+	cfg, err := config.Parse(configYamlPath)
```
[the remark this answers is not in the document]
adamkpickering: See https://github.com/rancher/image-mirror/pull/829#discussion_r1980265669

main.go
```diff
@@ -0,0 +1,98 @@
...
+			return nil, fmt.Errorf("source image split into %d parts (>=2 parts expected)", len(parts))
+		}
+		repoName := parts[len(parts)-2]
+		imageName := parts[len(parts)-1]
+		targetImageName = "mirrored-" + repoName + "-" + imageName
+	}
+
+	entries := make([]regsync.ConfigSync, 0, len(image.Tags))
+	for _, tag := range image.Tags {
+		sourceImage := image.SourceImage + ":" + tag
```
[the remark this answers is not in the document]
adamkpickering: See https://github.com/rancher/image-mirror/pull/829#discussion_r1980265669

main.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func getRegsyncEntries(repo config.Repository, image config.Image) ([]regsync.ConfigSync, error) {
+	targetImageName := image.TargetImageName
+	if targetImageName == "" {
+		parts := strings.Split(image.SourceImage, "/")
+		if len(parts) < 2 {
+			return nil, fmt.Errorf("source image split into %d parts (>=2 parts expected)", len(parts))
+		}
+		repoName := parts[len(parts)-2]
```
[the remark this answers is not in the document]
adamkpickering: Funny enough I considered using the equivalent from `regclient/regclient`. Thank you for letting me know that this function exists - it looks much better than the `regclient/regclient` one. I think that eventually these scripts will use a function like this.

However, at this time, we don't need it. In fact, using a function like this would complicate the code: we need to construct the target image name from the last two parts of the "image path". For example, if the source image is `rancher/fleet`, the default target image name would be `mirrored-rancher-fleet`. If we use `registry.ParseReference(url)` as you suggest, the "image path" is represented by the `Reference.Repository` field, which would still need to be split as we are currently doing.

I fully expect this code to evolve as we use it. I only intend this PR as an MVP - we will add more stuff like this as we go on.

main_test.go
diogoasouza: I think it's fine that it is in another PR

main.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/urfave/cli/v2"
+)
+
+const configYamlPath = "config.yaml"
+const regsyncYamlPath = "regsync.yaml"
+
+func main() {
+	log.SetFlags(0)
+
+	app := &cli.App{
```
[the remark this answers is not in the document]
brandond: I am ambivalent on the requested change. It is acceptable to just do `&cli.App{}` as seen here if you don't need or want any of the app defaults. There is no consistently used pattern across projects. Ref:
* https://github.com/k3s-io/helm-controller/blob/master/main.go#L19
* https://github.com/rancher/system-upgrade-controller/blob/master/main.go#L31

main_test.go
adamkpickering: I changed this to what you suggested in our slack conversation. Please resolve this if you're happy with it.

main.go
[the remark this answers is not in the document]
adamkpickering: I am ambivalent as well. I've changed it as requested.

brandond (approved):

rohitsakala (approved):

diogoasouza (approved):

PR #854, opened 2025-03-18
Add github actions workflow for mirroring images using regsync

adamkpickering:
Also:
- Add a workflow for running `golangci-lint` and `go test` on Go code during PR checks.
- Simplify user/password/registry settings in `config.yaml`.

Until we are more confident in these changes, regsync-based mirroring can only be triggered via a `workflow_dispatch` event.

### DIFF
--- a/.github/workflows/mirror-images-regsync.yml
+++ b/.github/workflows/mirror-images-regsync.yml
@@ -13,8 +13,28 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Read Secrets
+      - name: Check for secrets being set by workflow_dispatch
+        id: check_manual_secrets
+        run: |
+          if [ -n "${{ secrets.DEBUG_DOCKER_USERNAME }}" ]; then
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          elif [ -n "${{ secrets.DEBUG_DOCKER_PASSWORD }}" ]; then
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          elif [ -n "${{ secrets.DEBUG_APPCO_USERNAME }}" ]; then
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          elif [ -n "${{ secrets.DEBUG_APPCO_PASSWORD }}" ]; then
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          elif [ -n "${{ secrets.DEBUG_PRIME_USERNAME }}" ]; then
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          elif [ -n "${{ secrets.DEBUG_PRIME_PASSWORD }}" ]; then
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          else
+            echo "manual_secrets=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Read Secrets (only if not set in workflow_dispatch)
         uses: rancher-eio/read-vault-secrets@main
+        if: ${{ steps.check_manual_secrets.outputs.manual_secrets == 'true' }}
         with:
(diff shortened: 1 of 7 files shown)


diogoasouza (dismissed):
LGTM

.github/workflows/mirror-images-regsync.yml
```diff
@@ -0,0 +1,64 @@
...
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          elif [ -n "${{ secrets.DEBUG_APPCO_PASSWORD }}" ]; then
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          elif [ -n "${{ secrets.DEBUG_PRIME_USERNAME }}" ]; then
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          elif [ -n "${{ secrets.DEBUG_PRIME_PASSWORD }}" ]; then
+            echo "manual_secrets=true" >> $GITHUB_OUTPUT
+          else
+            echo "manual_secrets=false" >> $GITHUB_OUTPUT
+          fi
```
brandond, line 33: is this something that needs to be committed, or is this just for debugging in PR CI?

.github/workflows/mirror-images-regsync.yml
```diff
@@ -0,0 +1,64 @@
...
+        env:
+          # We use a ternary operator here. See
+          # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example
+          # for more details.
+          DOCKER_USERNAME: ${{ secrets.DEBUG_DOCKER_USERNAME && secrets.DEBUG_DOCKER_USERNAME || secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DEBUG_DOCKER_PASSWORD && secrets.DEBUG_DOCKER_PASSWORD || secrets.DOCKER_PASSWORD }}
+          APPCO_USERNAME: ${{ secrets.DEBUG_APPCO_USERNAME && secrets.DEBUG_APPCO_USERNAME || secrets.APPCO_USERNAME }}
+          APPCO_PASSWORD: ${{ secrets.DEBUG_APPCO_PASSWORD && secrets.DEBUG_APPCO_PASSWORD || secrets.APPCO_PASSWORD }}
+          PRIME_USERNAME: ${{ secrets.DEBUG_PRIME_USERNAME && secrets.DEBUG_PRIME_USERNAME || secrets.PRIME_USERNAME }}
+          PRIME_PASSWORD: ${{ secrets.DEBUG_PRIME_PASSWORD && secrets.DEBUG_PRIME_PASSWORD || secrets.PRIME_PASSWORD }}
```
brandond, line 64: same question - what is the purposed of the DEBUG_ secrets?

.github/workflows/mirror-images-regsync.yml
adamkpickering: This is for developing the workflow. Specifically, its for setting the secrets when running the workflow in one's personal fork - there is no way to do this when fetching the secrets via the EIO action.

Do you not want it there? I suppose this can be added back if we need to develop further, but IMO it's nice to have for when things break in the future. Certainly not ideal, but then again, neither is github actions...

.github/workflows/mirror-images-regsync.yml
adamkpickering: Addressed in https://github.com/rancher/image-mirror/pull/854#discussion_r2017533560

.github/workflows/mirror-images-regsync.yml
brandond: I'm a little sketched out by having stuff that's not used by GHA, in the GHA config. Is there any reason you can't just set the non-DEBUG env vars when you're running the tests locally?

brandond (approved):

PR #875, opened 2025-04-16
Introduce code to automate migration of `images-list` and `images-list-daily`

adamkpickering:
This PR is the first step towards https://github.com/rancher/rancher/issues/49869. It automates the process of migrating lines from `images-list` and `images-list-daily` to `config.yaml`.

This was an old branch that I wrote a month or two ago. At that time, my intent was to migrate all of the images in one go. However, it has become clear that we need to take a more granular approach. I've added a couple of commits onto the branch in order to facilitate this; `bin/image-mirror-tools migrate-images-list` now requires the source and target image (as specified in `images-list`) to be passed as arguments. Sorry for the confusion this may create when reviewing this PR.

### DIFF
--- a/tools/main.go
+++ b/tools/main.go
@@ -34,6 +34,11 @@ func main() {
 		},
 		Commands: []*cli.Command{
 			{
+				Name:   "format",
+				Usage:  "Enforce formatting on certain files",
+				Action: formatFiles,
+			},
+			{
 				Name:   "generate-regsync",
 				Usage:  "Generate regsync.yaml",
 				Action: generateRegsyncYaml,
@@ -189,4 +194,15 @@ func convertImageListEntryToImage(imageListEntry legacy.ImagesListEntry) (*confi
 	image.SetTargetImageName(targetImageName)
 
 	return image, nil
+}
+
+func formatFiles(_ context.Context, _ *cli.Command) error {
+	configJson, err := config.Parse(configYamlPath)
+	if err != nil {
+		return fmt.Errorf("failed to parse config: %w", err)
+	}
+	if err := config.Write(configYamlPath, configJson); err != nil {
+		return fmt.Errorf("failed to write config: %w", err)
+	}
+	return nil
 }
--- a/tools/main.go
+++ b/tools/main.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"strings"
 
 	"github.com/rancher/image-mirror/pkg/config"
 	"github.com/rancher/image-mirror/pkg/regsync"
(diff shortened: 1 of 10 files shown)


tools/pkg/config/config.go
```diff
@@ -57,5 +61,37 @@ func Parse(fileName string) (Config, error) {
...
+	if len(parts) < 2 {
+		return fmt.Errorf("source image split into %d parts (>=2 parts expected)", len(parts))
+	}
+	repoName := parts[len(parts)-2]
+	imageName := parts[len(parts)-1]
+	image.defaultTargetImageName = "mirrored-" + repoName + "-" + imageName
+	return nil
+}
+
+func (image Image) TargetImageName() string {
```
diogoasouza, line 134: I think this should be `func (image *Image)` to not potentially cause confusion by mixing methods with pointer and value receivers

tools/main.go
```diff
@@ -100,3 +108,65 @@ func convertConfigImageToRegsyncImages(repo config.Repository, image *config.Ima
...
+		return fmt.Errorf("failed to parse %q: %w", imagesListPath, err)
+	}
+
+	configJson, err := legacy.ParseConfig(configJsonPath)
+	if err != nil {
+		return fmt.Errorf("failed to parse %q: %w", configJsonPath, err)
+	}
+
+	for _, legacyImage := range legacyImages {
+		// if image in config.json, skip
```
diogoasouza: Noob question, but why are we skipping here?

tools/main.go
```diff
@@ -100,3 +108,65 @@ func convertConfigImageToRegsyncImages(repo config.Repository, image *config.Ima
...
+	if err := config.Write(configYamlPath, configYaml); err != nil {
+		return fmt.Errorf("failed to write %s: %w", configYamlPath, err)
+	}
+
+	return nil
+}
+
+func convertImageListEntryToImage(imageListEntry legacy.ImagesListEntry) (*config.Image, error) {
+	parts := strings.Split(imageListEntry.Target, "/")
+	if len(parts) < 2 {
```
diogoasouza, line 185: `if len(parts) != 2`

tools/pkg/config/config.go
```diff
@@ -14,6 +14,7 @@ type Config struct {
 	Repositories []Repository
 }
 
+// Image should not be instantiated directly. Instead, use NewImage().
```
diogoasouza, line 17: Shouldn't it be private in this case? I.e. `image` instead of `Image`?

tools/main.go
```diff
@@ -110,33 +119,39 @@ func convertConfigImageToRegsyncImages(repo config.Repository, image *config.Ima
...
 
+	if configJson.Contains(sourceImage) {
+		log.Printf("warning: %s refers to image with source %q", configJsonPath, sourceImage)
+	}
+
+	legacyImagesToKeep := make([]legacy.ImagesListEntry, 0, len(legacyImages))
 	for _, legacyImage := range legacyImages {
-		// if image in config.json, skip
-		if configJson.Contains(legacyImage.Source) {
+		if legacyImage.Source != sourceImage || legacyImage.Target != targetImage {
```
diogoasouza, line 158: I think this would be clearer without the negatives
```
if legacyImage.Source == sourceImage && legacyImage.Target == targetImage {
  newImage, err := convertImageListEntryToImage(legacyImage)
  if err != nil {
    return fmt.Errorf("failed to convert %q: %w", legacyImage, err)
  }
  accumulator.AddImage(newImage)
} else {
  legacyImagesToKeep = append(legacyImagesToKeep, legacyImage)
  continue
}
```

tools/main.go
```diff
@@ -110,33 +119,39 @@ func convertConfigImageToRegsyncImages(repo config.Repository, image *config.Ima
...
+		return fmt.Errorf("failed to parse images list: %w", err)
 	}
 
 	configJson, err := legacy.ParseConfig(configJsonPath)
 	if err != nil {
 		return fmt.Errorf("failed to parse %q: %w", configJsonPath, err)
 	}
 
+	if configJson.Contains(sourceImage) {
+		log.Printf("warning: %s refers to image with source %q", configJsonPath, sourceImage)
```
diogoasouza, line 153: `log.Printf("warning: %s refers to image with source %q \n", configJsonPath, sourceImage)`

tools/pkg/config/config.go
adamkpickering: This always trips me up. https://go.dev/doc/faq#methods_on_values_or_pointers says:

> Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used.

Which is good enough for me :+1:

tools/pkg/config/config.go
adamkpickering: I don't think so - we need to be able to pass `Image`s to functions, and return them from functions (for example, `convertImageListEntryToImage()`), which requires that the type is public.

tools/main.go
adamkpickering: The newline isn't necessary when using `log.Printf()` - the logging package ensures every log statement is on a separate line. See https://go.dev/play/p/T8tbtUuB7Tb

adamkpickering: Actually, I'm simply going to stop using the `log` package.

diogoasouza (approved):

PR #889, opened 2025-04-24
Get secrets from `env` context, not `secrets` context

adamkpickering:
Looking at https://github.com/rancher-eio/read-vault-secrets, it looks like the secrets go into the `env` context, not the `secrets` context. This PR fixes this for the regsync image mirroring workflow.

### DIFF
--- a/.github/workflows/mirror-images-regsync.yml
+++ b/.github/workflows/mirror-images-regsync.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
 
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror-images-regsync.yml
```diff
@@ -34,9 +34,9 @@ jobs:
           export PATH=$PATH:$(pwd)
           time regsync once --missing --config regsync.yaml
         env:
```
mallardduck, line 36: Technically I think you can remove all these...since it's already in the global env after the EIO read step. This doesn't break anything but is likely redundant unless renaming ENVs for this particular step.

mallardduck (approved):

PR #918, opened 2025-05-08
Migrate `images-list` to `config.yaml` (I-K)

adamkpickering:
This issue is part of https://github.com/rancher/rancher/issues/49870.

It migrates entries from `images-list` with source images starting with letters I-K that are not mentioned in `retrieve-image-tags/config.json`.


config.yaml
```diff
@@ -409,6 +409,305 @@ Images:
...
+- SourceImage: istio/sidecar_injector
+  Tags:
+  - 1.5.9
+- SourceImage: jaegertracing/all-in-one
+  Tags:
+  - 1.20.0
+  TargetImageName: jaegertracing-all-in-one
+- SourceImage: jaegertracing/all-in-one
+  Tags:
+  - "1.14"
```
diogoasouza, line 655: Why the quotes here?

adamkpickering: I think it's because it would be interpreted as a number without the quotes. See https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml:

> Use quotes to force a string, e.g. if your key or value is 10 but you want it to return a String and not a Fixnum, write '10' or "10".

I guess that the library I'm using for yaml decided they need to be there. Either way, this doesn't cause any issues.

diogoasouza (approved):

PR #919, opened 2025-05-09
Migrate `images-list` to `config.yaml` (L-L)

adamkpickering:
This issue is part of https://github.com/rancher/rancher/issues/49870.

It migrates entries from `images-list` with source images starting with letters L-L that are not mentioned in `retrieve-image-tags/config.json`.


diogoasouza (approved):
lgtm

PR #933, opened 2025-05-21
Fix `ImageAccumulator` indexing

adamkpickering:
Previously, `ImageAccumulator` indexed only on `SourceImage` and `TargetImageName`. This could lead to problems in cases like the one introduced in https://github.com/rancher/image-mirror/pull/932. This PR adds `DoNotMirror` to the indexed fields.


mallardduck (approved):
lgtm

PR #962, opened 2025-06-13
Adding Elemental 1.6.9 images to new Elemental repos

fgiudici:
registry.suse.com/rancher/elemental-operator:1.6.9 registry.suse.com/rancher/seedimage-builder:1.6.9

to be cloned on docker hub only keeping the same image name

#### Additional Notes ####

This is the first image sync requested for the new dockerhub rancher/elemental-operator and rancher/seedimage-builder images (https://github.com/rancherlabs/eio/issues/3418 , https://github.com/rancherlabs/eio/issues/3419).

This will validate the new repo and prepare for forthcoming Elemental releases where we will just need to sync images to docker hub (Elemental images are natively built in registry.suse.com) and stop mirroring the images to registry.suse.com.

The new repos anyway are not being used at the moment for the Elemental Rancher charts.
At repo switch time we will take care to update the Rancher's `origins.go` files accordingly.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1637,6 +1637,10 @@ Images:
   - 15.6.19.1
   - 15.6.24.2
   TargetImageName: mirrored-bci-micro
+- SourceImage: registry.suse.com/rancher/elemental-operator
+  Tags:
+  - 1.6.9
+  TargetImageName: elemental-operator
 - DoNotMirror:
   - 1.3.4
   SourceImage: registry.suse.com/rancher/elemental-operator
@@ -1663,6 +1667,10 @@ Images:
   - 1.6.8
   - 1.6.9
   TargetImageName: mirrored-elemental-seedimage-builder
+- SourceImage: registry.suse.com/rancher/seedimage-builder
+  Tags:
+  - 1.6.9
+  TargetImageName: seedimage-builder
 - SourceImage: rpardini/docker-registry-proxy
   Tags:
   - 0.6.1
--- a/regsync.yaml
+++ b/regsync.yaml
@@ -7173,6 +7173,9 @@ sync:
 - source: registry.suse.com/bci/bci-micro:15.6.24.2
   target: registry.suse.com/rancher/mirrored-bci-micro:15.6.24.2
   type: image
+- source: registry.suse.com/rancher/elemental-operator:1.6.9
+  target: docker.io/rancher/elemental-operator:1.6.9
+  type: image
 - source: registry.suse.com/rancher/elemental-operator:1.4.2
   target: docker.io/rancher/mirrored-elemental-operator:1.4.2
   type: image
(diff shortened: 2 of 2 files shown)


fgiudici:
Pull request checks failure due to https://github.com/rancher/image-mirror/pull/963

config.yaml
```diff
@@ -1147,6 +1147,10 @@ Images:
   - 15.6.19.1
   - 15.6.24.2
   TargetImageName: mirrored-bci-micro
+- SourceImage: registry.suse.com/rancher/elemental-operator
+  Tags:
+  - 1.6.9
+  TargetImageName: elemental-operator
```
brandond, line 1153: if we're adding something new, shouldn't it have the mirrored- prefix?

adamkpickering: @fgiudici and I talked about this over slack. TLDR: he wants to mirror `registry.suse.com/rancher/elemental-operator` to `docker.io/rancher/elemental-operator` for simplicity's sake. Otherwise we are mirroring it to `docker.io/rancher/mirrored-elemental-operator` and `registry.suse.com/rancher/mirrored-elemental-operator`, which can be confusing because we are mirroring the image back to the registry it came from. Whether the `mirrored-` prefix applies is unclear, since we produce `elemental-operator` internally. AFAIK the current setup is fine, but I might not know something @fgiudici does.

Let me know if you want more of the history of the elemental images + image-mirror.

config.yaml
```diff
@@ -1175,6 +1179,10 @@ Images:
   - 1.6.8
   - 1.6.9
   TargetImageName: mirrored-elemental-seedimage-builder
+- SourceImage: registry.suse.com/rancher/seedimage-builder
+  Tags:
+  - 1.6.9
+  TargetImageName: seedimage-builder
```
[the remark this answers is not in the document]
adamkpickering: NB `seedimage-builder` images are built and mirrored like the `elemental-operator` images.

adamkpickering (approved):
LGTM, but let's wait for @brandond's approval to merge.

brandond (approved):
Originally we'd decided that anything that isn't pushed directly to docker hub would have a mirrored- prefix, to reflect that it is managed by the mirror script, and docker hub is not the primary source for the image. This would include things that are built internally and primarily live on the suse registry.

If we are going to change that policy that's fine, I just wanted to make sure we have a discussion first.

cc @snasovich @Jono-SUSE-Rancher

brandond:
Yes, that was my point. If it's built and pushed directly to the rancher org it should not have the mirrored prefix. If it is built and pushed somewhere else, and then mirrored here, it should have the mirrored prefix. The org contains a mix of both and without adhering to this convention it is hard to tell what is what.

fgiudici:
> Pushing the images by ourselves to dockerhub and the prime registry sounds good to me. That way we can keep image-mirror out of the picture completely, except for the existing entries in `config.yaml` of course.
> 
> I thought that pushing to dockerhub was not possible with IBS, but I must have misunderstood. Closing this PR sounds good to me 👍

Yeah, I don't think it is possible to push to Docker Hub from IBS. 
I now understand the reason of a "mirrored-" prefix in images when dealing with synced images via image-mirror.
Since Elemental already builds the images in IBS and has them released in registry.suse.com/rancher it should be on Elemental to publish on the Docker Hub directly too (probably the build from upstream/OBS happening before the IBS one).

Currently we will keep using the mirrored- elemental images as usual.

PR #970, opened 2025-06-18
Allow `DoNotMirror` to be string slice

adamkpickering:
Issue: https://github.com/rancher/rancher/issues/50476


tools/internal/config/image.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+			if blacklistedTag == tag {
+				return true, nil
+			}
+		}
+		return false, nil
+	default:
+		return false, nil
+	}
+}
```
brandond, line 121: This will be O(Tag^DoNotMirror).

Might I suggest that instead of type-checking and iterating over the doNotMirror tag list for each tag, you add a couple fields to the Image struct with `https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets`:
```golang
type Image struct {
	...
	excludedTags   sets.Set[string]
	excludeAllTags bool
}
```

Populate those fields using this same logic in `NewImage`, and then `doNotMirrorTag()` simply becomes `return image.excludeAllTags || image.excludedTags(tag)`

brandond (changes requested):

tools/internal/config/config.go
```diff
@@ -145,32 +127,6 @@ func (config *Config) ToRegsyncConfig() (regsync.Config, error) {
...
-			Target: targetImage,
-			Type:   "image",
-		}
-		entries = append(entries, entry)
-	}
-
-	return entries, nil
-}
-
 func CompareImages(a, b *Image) int {
```
diogoasouza, line 145: Can we move this to the `image.go` file too?

tools/internal/config/image.go
adamkpickering: I made your suggested change, except using `map[string]struct{}` for sets - I didn't want the dependency. Let me know what you think!

tools/internal/config/image.go
```diff
@@ -0,0 +1,139 @@
...
+
+// ToRegsyncImages converts image into one ConfigSync (i.e. an image
+// for regsync to sync) for each tag present in image. repo provides
+// the target repository for each ConfigSync.
+func (image *Image) ToRegsyncImages(repo Repository) ([]regsync.ConfigSync, error) {
+	entries := make([]regsync.ConfigSync, 0, len(image.Tags))
+	for _, tag := range image.Tags {
+		if _, excluded := image.excludedTags[tag]; excluded || image.excludeAllTags {
+			continue
+		}
```
brandond, line 127: Why bother checking all the tags if we know they're all excluded?
```suggestion
	if image.excludeAllTags {
		return entries, nil
	}
	for _, tag := range image.Tags {
		if _, excluded := image.excludedTags[tag]; excluded {
			continue
		}
```

config.yaml
```diff
@@ -1164,8 +1164,11 @@ Images:
   - 15.6.19.1
   - 15.6.24.2
   TargetImageName: mirrored-bci-micro
-- SourceImage: registry.suse.com/rancher/elemental-operator
+- DoNotMirror:
```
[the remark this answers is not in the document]
adamkpickering: Good catch, not sure how this stayed in :+1:

tools/internal/config/config.go
adamkpickering: Funny enough, I originally moved it, but I moved it back. I don't feel strongly either way though - I'll move it to `image.go` as you suggest.

brandond (approved):

PR #991, opened 2025-06-25
Issue 49873

diogoasouza:
Issue https://github.com/rancher/rancher/issues/49873

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,19 +71,22 @@ entry specifies a strategy for finding tags of images to potentially add to
 | Field | Required | Description |
 | ------------- | ------------- |------------- |
 | `Name` | yes | A unique identifier for this autoupdate entry. Used for logging and generating branch names for pull requests.
-| `GithubLatestRelease` | no | See [`GithubLatestRelease`](#githublatestrelease).
+| `GithubRelease` | no | See [`GithubRelease`](#githubrelease).
 | `HelmLatest` | no | See [`HelmLatest`](#helmlatest).
 
-#### `GithubLatestRelease`
+#### `GithubRelease`
 
-The `GithubLatestRelease` strategy fetches the latest release tag from a GitHub
-repository and applies it to the specified images.
+The `GithubRelease` strategy fetches all release tags that matches the VersionConstraint from a GitHub
+repository and applies it to the specified images. 
+If LatestOnly is true, it only fetches from the latest release and does not consider the VersionConstraint.
 
-| Field | Required | Description |
-| ------------- | ------------- |------------- |
-| `Owner` | yes | The GitHub repository owner/organization.
-| `Repository` | yes | The GitHub repository name.
-| `Images` | yes | See [`Images`](#images).
+| Field               | Required | Description |
+|---------------------|----------|------------- |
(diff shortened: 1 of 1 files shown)


tools/internal/autoupdate/githublatestrelease.go
```diff
@@ -5,31 +5,90 @@ import (
...
+			if release.GetPrerelease() || release.GetDraft() {
+				continue
+			}
+			if gr.VersionConstraint != "" {
+				version, err := semver.NewVersion(release.GetTagName())
+				if err != nil {
+					return nil, fmt.Errorf("error parsing release version: %w", err)
+				}
+
+				constraint, err := semver.NewConstraint(gr.VersionConstraint)
```
adamkpickering: nit: since there are other comments we should address, we might as well put this outside of the nested for loop.

tools/internal/autoupdate/githublatestrelease.go
```diff
@@ -5,31 +5,90 @@ import (
...
+					return nil, fmt.Errorf("error parsing version constraint: %w", err)
+				}
+
+				//check if tag matches the constraint
+				if !constraint.Check(version) {
+					continue
+				}
+			}
+			for _, sourceImage := range gr.Images {
+				image, err := config.NewImage(sourceImage.SourceImage, []string{release.GetTagName()})
```
adamkpickering: nit: it would be nice to do some deduplication of `config.Image`s. As it is, if we find multiple releases that match the version constraint, we will get multiple `config.Image`s with one tag each. It's not a functional problem, so feel free to ignore this.

tools/internal/autoupdate/githublatestrelease.go
```diff
@@ -5,31 +5,90 @@ import (
...
+// GithubRelease retrieves the tags of all github releases that match the VersionConstraint if
+// LatestOnly is false, or the tag of the latest release if LatestOnly is true.
+// It returns the configured Images with this tag.
+// It assumes that the images have the same tags as the github releases.
+type GithubRelease struct {
+	Owner             string
+	Repository        string
+	Images            []AutoupdateImageRef
+	LatestOnly        bool
+	VersionConstraint string
```
adamkpickering: IMO we should cut down on visual noise by adding `omitempty` to both `LatestOnly` and `VersionConstraint`.

autoupdate.yaml
```diff
@@ -51,84 +53,118 @@
...
     - SourceImage: registry.k8s.io/sig-storage/csi-snapshotter
+    LatestOnly: true
     Owner: kubernetes-csi
     Repository: external-snapshotter
+    VersionConstraint: ""
   Name: csi-snapshotter
-- GithubLatestRelease:
+- GithubRelease:
+    Images:
+    - SourceImage: quay.io/coreos/etc
```
adamkpickering, line 98: Should this be `quay.io/coreos/etcd`?

adamkpickering:
Oh, also, we should update `README.md` with the new fields and `GithubLatestRelease` -> `GithubRelease` name change.

adamkpickering (approved):

PR #992, opened 2025-06-25
Add `HelmLatest` update strategy and migrate `calico-operator`

adamkpickering:
This is part of https://github.com/rancher/rancher/issues/50479.

This implementation of `HelmLatest` loosely follows the `helm-latest` parts of the python script. I initially planned on bug-for-bug compatibility, but the python script was a mess and the entire `helm-latest` update strategy needed rethinking.

TLDR: `HelmLatest` exclusively finds images referenced by a helm chart by running `helm template` with all of the configured sets of helm values, and looking for the `image` key in the yaml. There are too many edge cases to do more. I do not recommend the use of the `HelmLatest` update strategy, but include it for legacy reasons.

Determining which images are referenced by a helm chart is not an easy task. The python code takes two approaches:
1. Run `helm template` with one or more sets of helm values, and look for a key that contains (without respect to case) `image`.
2. Run `helm show values`, look for keys equal to `image`, and use `registry`, `repository` and `tag` inside the values of those keys.

2 is highly problematic. The majority of the charts I looked at had sections in `values.yaml` that defined `repository`, `tag` and maybe `registry`, but those sections were under keys other than `image`. Also, the python code has a bug here: it only goes through this process for the last chart configured in the given entry in `config.json`.

There are a number of other places that images could be defined/used:

- Operator-specific code and configuration, if the chart deploys an operator.
- Helm charts that depend on the main chart.
- CRDs.

It is infeasible to try to cover all of these cases. Instead, I have taken the simplest approach, which covers most cases. If users have issues I would likely suggest they use a more reliable update strategy.


tools/internal/autoupdate/config.go
```diff
@@ -77,10 +86,18 @@ func (entry ConfigEntry) Validate() error {
 		return errors.New("must specify Name")
 	}
 
-	if entry.GithubLatestRelease == nil {
+	if entry.GithubLatestRelease == nil && entry.HelmLatest == nil {
 		return errors.New("must specify an autoupdate strategy")
-	} else if err := entry.GithubLatestRelease.Validate(); err != nil {
-		return fmt.Errorf("GithubLatestRelease failed validation: %w", err)
+	} else if entry.GithubLatestRelease != nil && entry.HelmLatest != nil {
```
diogoasouza, line 91: I think it's clearer if we just have multiple `if` instead of `else if` since there's no chance of entering more than one if anyways

tools/internal/autoupdate/helmlatest.go
```diff
@@ -0,0 +1,200 @@
...
+				if err := hl.extractImagesFromYaml(value, imageMap); err != nil {
+					return err
+				}
+			}
+		}
+	case []any:
+		for _, item := range v {
+			if err := hl.extractImagesFromYaml(item, imageMap); err != nil {
+				return err
+			}
```
diogoasouza, line 159:
```suggestion
			return hl.extractImagesFromYaml(item, imageMap)
```

diogoasouza (approved):

PR #1001, opened 2025-07-03
Add registry strategy

diogoasouza:
Issue https://github.com/rancher/rancher/issues/50480


autoupdate.yaml
```diff
@@ -143,6 +148,18 @@
     Owner: kubernetes-csi
     Repository: livenessprobe
   Name: storage-livenessprobe
+- Name: traefik
+  Registry:
+    Images:
+    - SourceImage: library/traefik
+    Latest: true
+    VersionFilter: ^2.[0-9]+.[0-9]+$
+- Name: traefik3
```
brandond, line 157: Shouldn't need to split these out. Just join the regexes together.

diogoasouza: Makes sense 👍

tools/internal/autoupdate/registry.go
```diff
@@ -0,0 +1,348 @@
...
+			continue
+		}
+		filteredTags = append(filteredTags, tag)
+	}
+
+	if len(filteredTags) == 0 {
+		return nil, errors.New("no tags found matching version filter")
+	}
+
+	if r.LatestEntry {
```
adamkpickering, line 75: It might be a good idea to merge the `if` statements for `r.LatestEntry` and `r.Latest` into an `else-if` statement, since they are mutually exclusive. Feel free to disregard.

tools/internal/autoupdate/registry.go
```diff
@@ -0,0 +1,348 @@
...
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+	if token != "" {
+		req.Header.Set("Authorization", "Bearer "+token)
+	}
+	req.URL.RawQuery = params.Encode()
+	return req, nil
+}
+
+func (r *Registry) getAllFromImage() (registry string, namespace string, repository string) {
```
adamkpickering, line 189: This function name should be more descriptive of what the function does.

tools/internal/autoupdate/registry.go
```diff
@@ -0,0 +1,348 @@
...
+	"regexp"
+	"sort"
+	"strings"
+	"time"
+)
+
+type Registry struct {
+	Images        []AutoupdateImageRef
+	Latest        bool   `json:",omitempty"`
+	LatestEntry   bool   `json:",omitempty"`
```
adamkpickering, line 25: If you want, we could remove `LatestEntry`, since nothing uses it. Also, it is a potentially misleading name - it's hard to guarantee that the first tag is the latest one, given that we are supporting a number of different registries. Your call though.

tools/internal/autoupdate/registry.go
```diff
@@ -0,0 +1,348 @@
...
+		resp, err = httpClient.Do(req)
+		if err != nil {
+			return nil, fmt.Errorf("request failed: %w", err)
+		}
+
+		if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+			return resp, nil
+		}
+
+		if !retryableStatuses[resp.StatusCode] {
```
adamkpickering, line 256: It would be nice to print out more information than just the status code here.

tools/internal/autoupdate/registry.go
```diff
@@ -0,0 +1,348 @@
...
+			return nil, fmt.Errorf("request failed with status %s", resp.Status)
+		}
+
+		if resp.Body != nil {
+			io.Copy(io.Discard, resp.Body)
+			resp.Body.Close()
+		}
+
+		if i == maxRetries-1 {
```
adamkpickering, line 268: It seems strange to have this check here when we are using a `for` loop that checks `maxRetries` at the beginning of every loop. Can this be simplified?

README.md
```diff
@@ -111,6 +112,24 @@ output.
 | `Images` | no | Used to map a given update image to an entry in `config.yaml`. There may be multiple entries that have the same `SourceImage`, but different `TargetImageName`s, so we need to choose which one receives the update image.
 | `ImageDenylist` | no | A list of images to exclude from the results.
 
+#### `Registry`
+
+The `Registry` fetches all image tags that matches the VersionFilter from a registry defined in the Images provided.
```
adamkpickering, line 117: Nit: we could put `VersionFilter` in backticks so it appears as code.

adamkpickering: Also, did you mean "The `Registry` strategy fetches..."?

README.md
```diff
@@ -111,6 +112,24 @@ output.
...
+* K8s registry (registry.k8s.io)
+* GitHub Container Registry (ghcr.io)
+* Google Container Registry (gcr.io)
+
+| Field           | Required | Description |
+|-----------------|----------|------------- |
+| `Images`        | yes      | Used to map a given update image to an entry in `config.yaml`. There may be multiple entries that have the same `SourceImage`, but different `TargetImageName`s, so we need to choose which one receives the update image.
+| `Latest`        | no       | A flag to only use the latest tag.
+| `LatestEntry`   | no       | A flag to only use the latest entry in the registry.
+| `VersionFilter` | yes      | A regex to match against the image tags fetched from the registry.
```
adamkpickering, line 131: I don't think `VersionFilter` is required. At least, `Validate` doesn't require it.

README.md
```diff
@@ -111,6 +112,24 @@ output.
...
+* Docker Hub
+* Quay.io
+* K8s registry (registry.k8s.io)
+* GitHub Container Registry (ghcr.io)
+* Google Container Registry (gcr.io)
+
+| Field           | Required | Description |
+|-----------------|----------|------------- |
+| `Images`        | yes      | Used to map a given update image to an entry in `config.yaml`. There may be multiple entries that have the same `SourceImage`, but different `TargetImageName`s, so we need to choose which one receives the update image.
+| `Latest`        | no       | A flag to only use the latest tag.
```
adamkpickering, line 129: We should probably say something about it finding the latest tag via semver here. Not every tag is in semver format.

tools/internal/autoupdate/registry.go
```diff
@@ -0,0 +1,348 @@
...
+func parseLinkHeader(linkHeader string) string {
+	if linkHeader == "" {
+		return ""
+	}
+	link := strings.TrimPrefix(linkHeader, "<")
+	link = strings.TrimSuffix(link, `>; rel="next"`)
+	return link
+}
+
+func (r *Registry) processRegistryResponse(body []byte, linkHeader, githubToken, registry, namespace, pkg string, page int) ([]string, error) {
```
adamkpickering, line 290: The duplicated `switch-case` statements in `processRegistryResponse` and `buildRequest` are code smells. Arguably the best way to organize this is to give each registry its own type, and implement the code in each case inside that type.

On the other hand, this code is largely a translation of the original python. If you're feeling motivated, maybe consider restructuring this. Otherwise, feel free to disregard this comment.

adamkpickering (changes requested):

autoupdate.yaml
adamkpickering: I'm not sure this is a good idea. Merging the two entries, combined with using the `Latest` option, means that only the latest i.e. 3.x versions will get autoupdated. Traefik 2.11 is [supported until Feb. 2026](https://doc.traefik.io/traefik/deprecation/releases/). We want to mirror those 2.11 patch versions, right?

PR #1005, opened 2025-07-04
Fix `HelmLatest` update strategy's treatment of `Images` key

adamkpickering:
I noticed an error in the autoupdate workflow: https://github.com/rancher/image-mirror/actions/runs/16074545116/job/45366540428

We were treating `foundTargetImageName == ""` as an error case. However, this is also the case when `sourceImage == autoupdateImageRef.SourceImage` but `autoupdateImageRef.TargetImageName` is unset. This PR fixes this by making `foundTargetImageName` a pointer to a string, and treating the `nil` case as unset.

### DIFF
--- a/tools/internal/autoupdate/helmlatest.go
+++ b/tools/internal/autoupdate/helmlatest.go
@@ -97,22 +97,20 @@ func (hl *HelmLatest) GetUpdateImages() ([]*config.Image, error) {
 	// Convert the map to a slice of config.Image objects
 	images := make([]*config.Image, 0, len(imageMap))
 	for sourceImage, tags := range imageMap {
-		foundTargetImageName := ""
+		var foundTargetImageName *string
 		for _, autoupdateImageRef := range hl.Images {
 			if sourceImage == autoupdateImageRef.SourceImage {
-				foundTargetImageName = autoupdateImageRef.TargetImageName
+				foundTargetImageName = &autoupdateImageRef.SourceImage
 				break
 			}
 		}
-		if foundTargetImageName == "" {
+		if foundTargetImageName == nil {
 			return nil, fmt.Errorf("found image %s but it is not present in Images", sourceImage)
 		}
-
-		image, err := config.NewImage(sourceImage, tags, foundTargetImageName, nil)
+		image, err := config.NewImage(sourceImage, tags, *foundTargetImageName, nil)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


tools/internal/autoupdate/helmlatest.go
```diff
@@ -97,22 +97,20 @@ func (hl *HelmLatest) GetUpdateImages() ([]*config.Image, error) {
 	// Convert the map to a slice of config.Image objects
 	images := make([]*config.Image, 0, len(imageMap))
 	for sourceImage, tags := range imageMap {
-		foundTargetImageName := ""
+		var foundTargetImageName *string
 		for _, autoupdateImageRef := range hl.Images {
 			if sourceImage == autoupdateImageRef.SourceImage {
-				foundTargetImageName = autoupdateImageRef.TargetImageName
+				foundTargetImageName = &autoupdateImageRef.SourceImage
```
diogoasouza, line 103: I didn't understand this change, can you explain it? Why are we not using `autoupdateImageRef.TargetImageName`?

adamkpickering: There are three possibilities here:
1. We have found an image in the helm chart that is not mentioned in `Images`.
2. We found an image in the helm chart that matches `SourceImage` of an entry in `Images`, but that entry does not have `TargetImageName` defined.
3. We found an image in the helm chart that matches `SourceImage` of an entry in `Images`, and that entry has `TargetImageName` defined.

1 is an error condition. This is because we require the user to specify, for all images that might be found in the helm chart, what the `TargetImageName` value should be. We have this requirement because we don't want to add an image with a blank `TargetImageName` when there might already be one in `config.yaml` that has `TargetImageName` defined. Basically this must be an error because a human must make a choice here.

2 and 3 are easier, because we can map the `TargetImageName` (whether zero-valued or not) and `SourceImage` from `Images` directly to an entry in `config.yaml`.

It occurred to me that this might be a problem for the `GithubRelease` and `Registry` update strategies, but they use `Images` differently. They use it (its `SourceImage` keys, specifically) as something to append found tags to, whereas `HelmLatest` uses it to map found full images to `SourceImage`-`TargetImageName` pairs to add to in `config.yaml`.

diogoasouza: Got it, thanks!

diogoasouza (approved):

PR #1009, opened 2025-07-07
refactor registry logic

diogoasouza:


tools/internal/autoupdate/registry.go
```diff
@@ -118,118 +93,73 @@ func (r *Registry) Validate() error {
...
+		// Case 2: Handle images with a registry but no namespace like "k8s.gcr.io/pause"
+		registry = splittedImage[0]
+		namespace = ""
+		repository = splittedImage[1]
+	} else if len(splittedImage) > 3 && strings.Contains(splittedImage[0], ".") {
+		// Case 3: Handle images with long paths like "gcr.io/cloud-provider-vsphere/csi/release/syncer"
+		registry = splittedImage[0]
+		namespace = splittedImage[1]
+		repository = strings.Join(splittedImage[2:], "/")
+	} else {
```
adamkpickering, line 123: At this point, we haven't checked for the `len(splittedImage) == 1` case, which could cause a panic here.

tools/internal/autoupdate/registry.go
```diff
@@ -118,118 +93,73 @@ func (r *Registry) Validate() error {
...
+	if len(splittedImage) == 2 && !strings.Contains(splittedImage[0], ".") {
+		registry = "dockerhub"
+		namespace = splittedImage[0]
+		repository = splittedImage[1]
+	} else if len(splittedImage) == 2 && strings.Contains(splittedImage[0], ".") {
+		// Case 2: Handle images with a registry but no namespace like "k8s.gcr.io/pause"
+		registry = splittedImage[0]
+		namespace = ""
+		repository = splittedImage[1]
+	} else if len(splittedImage) > 3 && strings.Contains(splittedImage[0], ".") {
```
adamkpickering, line 118: I think that this case can be combined with the "standard 3-part" case. If the image has 3 parts, then `strings.Join(splittedImage[2:], "/")` should simply return `splittedImage[2]`.

diogoasouza: Good catch!

adamkpickering (approved):

PR #1029, opened 2025-07-14
Do not mirror old coreos images

adamkpickering:
These images are no longer accessible. Trying to pull them using `docker` results in a request for authentication. Trying to mirror them using `regsync` results in "unauthorized" errors (see [this workflow](https://github.com/rancher/image-mirror/actions/runs/16273748512/job/45947557578) for an example). This PR removes the problem images from mirroring, which should get the `regsync` workflow green again.


config.yaml
```diff
@@ -2204,23 +2204,28 @@ Images:
   - v0.14.0
   - v0.15.0
   - v0.15.1
-- SourceImage: quay.io/coreos/kube-state-metrics
+- DoNotMirror: true
```
brandond, line 2207: Maybe add a note as to why?
```suggestion
- DoNotMirror: true  # images have been removed from source
```

PR #1035, opened 2025-07-16
Migrate `images-list-daily`

adamkpickering:
Closes https://github.com/rancher/rancher/issues/50488.

My initial plan was as follows:
1. Put the images from `images-list-daily` into `config.yaml` and add a `Daily: true` field to them.
2. When writing out `regsync.yaml` shunt these images into `regsync-daily.yaml` based on the `Daily` field.
3. `regsync.yaml` and `regsync-daily.yaml` get their own workflows, similar to the current setup.

But, I realized that we aren't likely to update the `images-list-daily` images. Also there isn't much benefit to having them in `config.yaml`. They're a special, legacy case that breaks the assumption that is baked into `config.yaml` that tags are immutable. Accordingly I decided to omit steps 1 and 2, and simply manage `regsync-daily.yaml` manually.

If there are any problems with this, we can certainly look at doing it differently.

### DIFF
--- a/.github/workflows/mirror-images-daily.yml
+++ b/.github/workflows/mirror-images-daily.yml
@@ -39,7 +39,7 @@ jobs:
         run: |
           export PATH=$PATH:$(pwd)
           if [ "${{ inputs.debug }}" = "true" ]; then
-            time regsync once --logopt json --config regsync.yaml --verbosity debug
+            time regsync once --logopt json --config regsync-daily.yaml --verbosity debug
           else
-            time regsync once --logopt json --config regsync.yaml
+            time regsync once --logopt json --config regsync-daily.yaml
           fi


.github/workflows/mirror-images-daily.yml
```diff
@@ -1,48 +1,45 @@
...
+          curl --silent --fail --location --output regsync https://github.com/regclient/regclient/releases/download/v0.8.3/regsync-linux-amd64
+          chmod +x regsync
+
+      - name: Sync Container Images
+        run: |
+          export PATH=$PATH:$(pwd)
+          if [ "${{ inputs.debug }}" = "true" ]; then
+            time regsync once --logopt json --config regsync.yaml --verbosity debug
+          else
+            time regsync once --logopt json --config regsync.yaml
```
diogoasouza, line 44:
```suggestion
            time regsync once --logopt json --config regsync-daily.yaml --verbosity debug
          else
            time regsync once --logopt json --config regsync-daily.yaml
```

diogoasouza (changes requested):

diogoasouza (approved):
LGTM

PR #1087, opened 2025-09-05
Add PR check that ensures that the dockerhub repo exists for images added to `config.yaml`

adamkpickering:
Closes https://github.com/rancher/rancher/issues/51764.

Like the title says. Uses the dockerhub API to get the list of repositories under the `rancher` org, and then ensures that any images added relative to the merge base (default: `master`) are on this list.


diogoasouza (approved):
lgtm

PR #1130, opened 2025-10-23
Add `VersionRegex` field to `GithubRelease` autoupdate strategy

adamkpickering:
Closes https://github.com/rancher/rancher/issues/51451.

This PR adds the `VersionRegex` field to the `GithubRelease` autoupdate strategy. This field lets you filter github release tags using regex, as well as extract parts of the tags via the first capture group.


brandond (approved):
lgtm!

To be clear, this is hardcoded to use the first capture group from the regex as the value - if anyone needs to get fancy with it they should be able to just use nested groups I think.

brandond (approved):
🧨

PR #1142, opened 2025-10-30
Add `suse-virtual-cluster-engine:1.0.0`

enrichman:

- [ ] If an entire image is being added, a repo has been created for the image in DockerHub under the `rancher` org
- [ ] Additions are licensed with Rancher favored licenses - Apache 2 and MIT - or approved licenses - as according to [CNCF approved licenses](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md).
- [ ] Additions, when used in Rancher or Rancher's provided charts, have their corresponding origin added in Rancher's images [origins](https://github.com/rancher/rancher/blob/release/v2.7/pkg/image/origins.go) file (must be added for all Rancher versions `>= v2.7`).
- [ ] Any changes to scripting or CI config have been tested to the best of your ability

- [ ] Confirm that you can pull the mirrored images and tags from all target locations

### DIFF
--- a/regsync.yaml
+++ b/regsync.yaml
@@ -581,6 +581,18 @@ sync:
 - source: dp.apps.rancher.io/charts/suse-virtual-cluster-engine:1.0.0
   target: stgregistry.suse.com/rancher/charts/appco-suse-virtual-cluster-engine:1.0.0
   type: image
+- source: dp.apps.rancher.io/containers/k3k:1.0.0-1.1
+  target: registry.suse.com/rancher/appco-k3k:1.0.0-1.1
+  type: image
+- source: dp.apps.rancher.io/containers/k3k:1.0.0-1.1
+  target: stgregistry.suse.com/rancher/appco-k3k:1.0.0-1.1
+  type: image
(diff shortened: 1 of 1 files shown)


regsync.yaml
```diff
@@ -611,6 +611,12 @@ sync:
 - source: dp.apps.rancher.io/containers/redis:8.0.3-2.1
   target: stgregistry.suse.com/rancher/appco-redis:8.0.3-2.1
   type: image
+- source: dp.apps.rancher.io/containers/suse-virtual-cluster-engine:1.0.0
```
recena: This is not correct.

I encourage to use the [image-lock file](https://apps.rancher.io/artifacts/2a83ac621b987549ab6f40007bb6a1d355fe287137c11afb12e8b406e2fdee15) associated to the Helm chart.

For SUSE Virtual Cluster Engine `1.0.0`:

- `dp.apps.rancher.io/containers/k3k:1.0.0-1.1`
- `dp.apps.rancher.io/containers/k3k-kubelet:1.0.0-1.1`

enrichman:
Thanks @recena, I've updated the images to point to the right tag.

I've also added the charts mirroring, not sure if we need also those though.

PR #1172, opened 2025-11-27
Changing image references to artifact

diogoasouza:
https://github.com/rancher/rancher/issues/52259

### DIFF
--- a/regsync-daily.yaml
+++ b/regsync-daily.yaml
@@ -3,7 +3,7 @@ creds:
   registry: docker.io
   user: '{{ env "DOCKER_USERNAME" }}'
 defaults:
-  userAgent: rancher-image-mirror
+  userAgent: rancher-artifact-mirror
 sync:
 - source: neuvector/compliance-config:latest
   target: rancher/mirrored-neuvector-compliance-config:latest
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -1,4 +1,4 @@
-module github.com/rancher/image-mirror
+module github.com/rancher/artifact-mirror
 
 go 1.24.0
 
--- a/tools/internal/autoupdate/config.go
+++ b/tools/internal/autoupdate/config.go
@@ -12,10 +12,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/rancher/image-mirror/internal/config"
-	"github.com/rancher/image-mirror/internal/git"
-	"github.com/rancher/image-mirror/internal/paths"
-	"github.com/rancher/image-mirror/internal/regsync"
+	"github.com/rancher/artifact-mirror/internal/config"
+	"github.com/rancher/artifact-mirror/internal/git"
+	"github.com/rancher/artifact-mirror/internal/paths"
+	"github.com/rancher/artifact-mirror/internal/regsync"
 
 	"github.com/google/go-github/v79/github"
 	"sigs.k8s.io/yaml"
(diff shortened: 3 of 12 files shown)


README.md
```diff
@@ -8,11 +8,11 @@ places they need to be for Rancher and its associated projects to use them.
 `regsync` is used for mirroring images from one repository to another. `regsync`
```
adamkpickering: We should update the title of the readme too.

README.md
```diff
@@ -8,11 +8,11 @@ places they need to be for Rancher and its associated projects to use them.
...
 is configured via `regsync.yaml`. `regsync.yaml` is generated by running the
```
adamkpickering: We should have a little blurb that says something like "this used to be image-mirror; it was renamed to artifact-mirror because... see issue X for more details". Maybe use something like in https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts?

README.md
```diff
@@ -23,26 +23,26 @@ an update is found, the automation creates a pull request that a human user
 can then review and merge. See [`autoupdate.yaml`](#autoupdateyaml) for
```
adamkpickering: I guess the reference to "images" earlier in this paragraph will be updated when you do the autoupdate PR?

README.md
```diff
@@ -23,26 +23,26 @@ an update is found, the automation creates a pull request that a human user
...
 more information.
 
-### Adding New Images
+### Adding New Artifacts
 
-When adding new images to the repo, please indicate so in the pull request.
+When adding new artifacts to the repo, please indicate so in the pull request.
```
adamkpickering: It's a little unclear what is meant by "artifact" by itself. We can refer to "artifacts" in most cases, but it is important that we introduce the term as "OCI artifact". This is one example of where I'd use the term "OCI artifacts" (the first usage in this paragraph). Another place I'd use the term "OCI artifact" is in code comments.

I'll let this comment represent all of those cases. There is no clear rule for what should be "artifact" and what should be "OCI artifact", so I'll trust your taste as to which occurrences to modify.

README.md
adamkpickering: Also this line has an occurence of "images" in it.

README.md
```diff
@@ -69,30 +69,30 @@ information to be available to users.
...
 
 | Field | Required | Description |
 | ------------- | ------------- |------------- |
-| `BaseUrl` | yes | The base URL for the repository. Appending `/` plus an image name should be a valid image reference.
+| `BaseUrl` | yes | The base URL for the repository. Appending `/` plus an artifact name should be a valid artifact reference.
 | `Password` | yes | The password to use when authenticating against the registry. See [the regsync documentation](https://regclient.org/usage/regsync/) for more details.
 | `Registry` | yes | The registry URL. See [the regsync documentation](https://regclient.org/usage/regsync/) for more details.
 | `ReqConcurrent` | no | The number of concurrent requests that are made to this registry. See [the regsync documentation](https://regclient.org/usage/regsync/) for more details.
-| `DefaultTarget` | no | Whether the Repository is used as a target repository for a given Image when the `TargetRepositories` field of the Image is not set.
+| `DefaultTarget` | no | Whether the Repository is used as a target repository for a given artifact when the `TargetRepositories` field of the artifact is not set.
```
adamkpickering: nit: these `artifact` occurrences could be capitalized, and maybe put in backticks, because they are references to the `Artifact` type.

